PR #137, opened 2025-09-01
Update README.md

LinXunFeng:
Updated the docs link from incorrect <https://docs.termix.site/docs> to <https://docs.termix.site/install>


LukeGus:
Thanks for the catch, however I have already updated this in the dev-1,4.0 branch which is launching in the next hour hence why the link is currently wrong. Thanks again.

PR #424, opened 2025-10-21
Handle OIDC users during database import

nikolanovoselec:
## Summary
- let the SQLite import API detect OIDC-only accounts and unlock their data key via authenticateOIDCUser while keeping the password flow for local users
- update the admin UI to hide the password prompt for OIDC sessions and only send the password field when a local account performs the import
- document the new control flow in both backend and frontend code so future maintainers understand the split paths

## Testing
- docker build -f docker/Dockerfile -t termix-oidc-test .


LukeGus:
Thanks. I'll merge it for v1.8.0 releasing in a week or two.

PR #513, opened 2026-01-15
Enter Key for Quick Login

aditya-tawade:
# Overview

- [x] Added: While Login on Login Page, User has to manually press the button of login via Mouse, I have added funtionality of Enter Key where when user just need to username/password > 2FA and Just Enter


LukeGus:
LGTM. Thanks!

PR #521, opened 2026-01-20
fix: SSH key passphrase not passed for Docker and Tunnel

ZacharyZcR:
## Summary
Fix SSH key authentication with passphrase failing for Docker and Tunnel connections.

Fixes Termix-SSH/Support#366

## Problem
When using SSH key with passphrase, Docker and Tunnel connections fail with:
```
Error: Cannot parse privateKey: Encrypted private OpenSSH key detected, but no passphrase given
```

## Root Cause
Database field is `key_password` (snake_case), but code accessed `host.keyPassword` (camelCase).

## Fix
Added fallback to check both field names:
```javascript
// Before
keyPassword: host.keyPassword

// After  
keyPassword: host.key_password || host.keyPassword
```

## Files Changed
- `src/backend/ssh/docker.ts` - Docker SSH connection
- `src/backend/ssh/tunnel.ts` - Tunnel connections (source and endpoint hosts)

## Test Plan
- [ ] Create host with SSH key + passphrase
- [ ] Connect via Docker - should work
- [ ] Connect via Tunnel - should work


XtraLarge:
I use 1.11.0 and got the following error onnstart docker connection (ssh works well with key and password for key)
Connection error: Cannot parse privateKey: Encrypted private OpenSSH key detected, but no passphrase given

PR #535, opened 2026-01-27
fix: translation with region

TomyJan:
# Overview

Fixed: use `locale_with_underscore` to support all translations

# Changes Made

replace `two_letters_code` with `locale_with_underscore`

# Related Issues

https://github.com/Termix-SSH/Support/issues/442

# Checklist


LukeGus:
Since the filename would change, the `LanguageSwitcher.tsx` and the `i18n.ts` need to be updated to account for it.

TomyJan:
> Since the filename would change, the `LanguageSwitcher.tsx` and the `i18n.ts` need to be updated to account for it.由于文件名会发生变化，因此需要更新 `LanguageSwitcher.tsx` 和 `i18n.ts` 以加以说明。

done

TomyJan:
Since I don't have access to pull translations, I can't confirm if this is correct.

LukeGus:
🤦Right, sorry about that. If I had remembered that you had said that you did not have perms, I would have done it myself. I'll double-check it to ensure it's correct. Thanks so much!

EDIT: Nevermind, that was a different conversation where someone mentioned not having perms to check, sorry for any confusion.

PR #668, opened 2026-04-07
Fix WebSocket reconnection and add connection lost overlay

Xenthys:
# Overview

Fix WebSocket disconnections causing terminal tabs to be lost and replace the full-page reload with a non-destructive reconnection overlay.

- [x] Fixed: WebSocket code 1006 (abnormal closure) now triggers reconnection instead of giving up permanently
- [x] Fixed: Auth failure (code 1008) no longer reloads the entire page, destroying all open tabs
- [x] Fixed: Zombie WebSocket connections no longer silently eat the connection limit
- [x] Added: "Connection Lost" overlay with auto-reconnect backoff, Retry and Reload buttons
- [x] Updated: Max reconnect attempts raised from 3 to 8, connection limit from 3 to 10

# Changes Made

**Backend (`terminal.ts`):**
- Added pong timeout detection, connections that don't respond within 30s are terminated, freeing up slots
- Raised per-user WebSocket connection limit from 3 to 10 to handle reconnection races

**Frontend (`Terminal.tsx`, desktop + mobile):**
- Code 1006 now attempts reconnection when the session was previously connected (was blocked entirely before)
- Code 1008 shows an inline error instead of calling `window.location.reload()`
- Added client-side pong tracking to detect dead connections within 30s
- Removed unconditional `terminal.clear()` on close to preserve output during reconnection
- Raised max reconnect attempts from 3 to 8

**Frontend (`DesktopApp.tsx`):**
- Replaced the full `Dashboard` render on backend disconnect with a translucent overlay
- Overlay auto-reconnects with exponential backoff (2s → 30s cap)
- Retry button appears after 5 failed attempts, Reload button always available
- All tabs and terminal state preserved underneath

# Related Issues

- Related to Termix-SSH/Support#538
- Related to Termix-SSH/Support#542

# Checklist

### DIFF
--- a/src/backend/ssh/terminal.ts
+++ b/src/backend/ssh/terminal.ts
@@ -339,7 +339,7 @@ const wss = new WebSocketServer({
 
       const existingConnections = userConnections.get(payload.userId);
 
-      if (existingConnections && existingConnections.size >= 3) {
+      if (existingConnections && existingConnections.size >= 10) {
         return false;
       }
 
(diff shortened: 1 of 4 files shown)


LukeGus:
LGTM, thanks.

PR #673, opened 2026-04-09
fix: prevent server status failure from blocking host list loading

### LINKED ISSUE #571: [BUG] Counters (servers credentials tunnels) zero'ed after import
### Title

Counters show zeroes

### Platform

N/A

Docker

### Version

2.0.0

### Troubleshooting

- [x] I have read Known Issues (https://github.com/Termix-SSH/Support/issues/296)

### The Problem

After importing my configuration between instances I've noticed that all counters show 0 Servers, Tunnels and Credentials:

Image

In "Hosts manager" i see proper number:

Image

No errors in log
(same on http as in app)

### How to Reproduce

1. Import config
2. See zeroes

### Additional Context

In different issue it can be seen on screenshot it shows 0 too: 
https://github.com/Termix-SSH/Support/issues/565

ZacharyZcR:
## Summary
- Fixed host list failing to load when the server status API is unavailable
- Root cause: `getSSHHosts()` called `getAllServerStatuses()` without independent error handling — if the status endpoint failed (e.g. stats service down, network timeout), the entire host list fetch threw, causing dashboard counters to show 0 and host manager to fail
- Wrapped `getAllServerStatuses()` in its own try-catch so status failures are silently ignored (hosts display with "unknown" status instead of not displaying at all)
- Also added `Array.isArray()` guard on `hostsResponse.data` to prevent `e.map is not a function` crashes when the API returns unexpected data

## Related Issues
Closes Termix-SSH/Support#571
Closes Termix-SSH/Support#615

## Test plan
- [ ] Verify host list loads normally when all services are healthy
- [ ] Stop the stats service (port 30005) — host list should still load with "unknown" status
- [ ] Dashboard counters should show correct server/tunnel/credential counts even if stats service is down
- [ ] Host manager should function normally regardless of status API availability

### DIFF
--- a/src/ui/main-axios.ts
+++ b/src/ui/main-axios.ts
@@ -1006,10 +1006,16 @@ function handleApiError(error: unknown, operation: string): never {
 export async function getSSHHosts(): Promise<SSHHostWithStatus[]> {
   try {
     const hostsResponse = await sshHostApi.get("/db/host");
-    const hosts: SSHHost[] = hostsResponse.data;
+    const hosts: SSHHost[] = Array.isArray(hostsResponse.data)
+      ? hostsResponse.data
+      : [];
 
-    const statusesResponse = await getAllServerStatuses();
-    const statuses = statusesResponse || {};
+    let statuses: Record<number, ServerStatus> = {};
+    try {
+      statuses = (await getAllServerStatuses()) || {};
+    } catch {
+      // Status fetch failure should not prevent host list from loading
+    }
 
     return hosts.map((host) => ({
       ...host,


chaunt1:
hi @LukeGus please release this bug fix into 2.0.1 than 2.1.0 due to auto update docker image v2.0.0 is crashing host list load

LukeGus:
@chaunt1 v2.1.0 is nearing release, and it's about to go into beta testing. Instead of deploying this as 2.0.1, would you mind using the beta image? The Docker image tag is "dev-2.1.0" so you can pull and deploy that image, and you shouldn't have issues with it, but do keep in mind that it is a beta version, so make backups if you have important data. If you need the desktop files, they can be found here if you log into GitHub and scroll down: https://github.com/Termix-SSH/Termix/actions/runs/24378534607 (as of now, the build files are still being generated; check back within the next hour).

chaunt1:
@LukeGus oh yes i found the image now. will change to use the beta until the official release updated. thank a lot guys this project is very helpful for me

PR #675, opened 2026-04-09
fix: show server config on first launch instead of auto-selecting embedded

### LINKED ISSUE #594: [FEATURE] Connect Standalone Connector to Backend Server
### Title

Connect Standalone Connector to Backend Server

### Platform

Desktop app (Windows, Linux, and macOS)

### Is it related to an issue?

Maybe

### The Solution

I installed the 2.0 NSIS Windows client and was never prompted to connect to a server.  Local is fine for now but is there a way to have the client connect to the already existing backend server?

### Additional Context

_No response_

ZacharyZcR:
## Summary
- Fixed the Electron desktop app never showing the server configuration dialog on first launch
- Root cause: `checkServerConfig()` auto-selected the embedded local server when no config was stored (`config === null`), because `!config?.serverUrl` evaluates to `true` for `null` — so first-time users were silently put in local-only mode with no way to enter a remote server URL
- Added `config &&` check to only auto-connect to embedded when the user has **previously chosen** embedded mode (config exists with empty serverUrl)
- When choosing "Use Local Server", the choice is now persisted via `saveServerConfig({ serverUrl: "" })` so subsequent app restarts correctly auto-connect without showing the dialog again

## Related Issue
Closes Termix-SSH/Support#594

## Test plan
- [ ] Fresh install of Electron desktop app — should show server config dialog with both "Use Local Server" and URL input options
- [ ] Choose "Use Local Server" — app connects to embedded backend, dialog doesn't appear on next restart
- [ ] Choose remote server URL — app connects to remote, dialog doesn't appear on next restart
- [ ] Existing users who previously configured a remote server — no change in behavior
- [ ] Web browser users — no change (server config is Electron-only)

### DIFF
--- a/src/ui/desktop/authentication/Auth.tsx
+++ b/src/ui/desktop/authentication/Auth.tsx
@@ -31,6 +31,7 @@ import {
   getOIDCAuthorizeUrl,
   verifyTOTPLogin,
   getServerConfig,
+  saveServerConfig,
   isElectron,
   getEmbeddedServerStatus,
   isEmbeddedMode,
@@ -772,7 +773,7 @@ export function Auth({
             getEmbeddedServerStatus(),
           ]);
 
-          if (status?.embedded && status?.running && !config?.serverUrl) {
+          if (status?.embedded && status?.running && config && !config.serverUrl) {
             setCurrentServerUrl("");
             setShowServerConfig(false);
             return;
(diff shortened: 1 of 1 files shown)


mbaldwin1k:
Thanks for the update!  Are we able to select the remote backend as well as local or if we choose local we are stuck on local?

ZacharyZcR:
> Thanks for the update! Are we able to select the remote backend as well as local or if we choose local we are stuck on local?

You can switch between them freely, but only one at a time.

mbaldwin1k:
Perfect.  Thank you again.

PR #701, opened 2026-04-22
Improve client handling of network issues (activity/log, metrics, status)

Xenthys:
# Overview

Fix several 500/404/Network errors between the client and the server, and replace the blocking "Connection Lost" overlay with a non-intrusive "recovering" toast so the UI stays usable during transient backend hiccups.

Disclaimer: AI assistance (Claude Opus 4.7) was used to find cases I may have missed and perform a sanity check on this code. This PR has been fully reviewed by myself (a human), built and tested on my own instance before submission.

- [x] Added: progressive retry schedule for `GET /status` (2s/5s/8s timeouts with 3s/5s pauses, 23s total in the worst case). Fits inside one poll cycle so attempts don't stack.
- [x] Added: persistent, non-dismissible "Server connection lost, recovering…" toast carrying a Reload action; replaces the blocking overlay.
- [x] Added: `getServerMetricsById` now silently returns `null` on 404 instead of throwing, so the client stops warn-logging expected "no metrics yet / disabled" responses.
- [x] Updated: `POST /metrics/register-viewer` is a graceful 200 no-op when the host cannot be found, metrics are disabled, or the connection type does not support metrics (returns `{ success: true, skipped: true, reason }`).
- [x] Updated: `PollingManager.registerViewer` wraps the fire-and-forget `startMetricsForHost` with a `.catch` so a background start-up failure can no longer leak as a 500 on the register request.
- [x] Updated: Dashboard fetches `/status` once per refresh cycle and skips hosts known to be offline before asking for their metrics (unless `statusCheckEnabled === false`, in which case we can't know and still attempt).
- [x] Updated: Dashboard honors the new `skipped` flag from `registerMetricsViewer` and does not try to fetch metrics for those hosts.
- [x] Updated: `dbHealthMonitor.reportDatabaseError` now also triggers on `ECONNABORTED` / `ECONNRESET` / `ETIMEDOUT` / `ERR_CANCELED` and on "Request aborted" / timeout error messages, alongside the existing `ERR_NETWORK` / `ECONNREFUSED` and `database`/`drizzle`/`sqlite` patterns.
- [x] Removed: the blocking full-screen `ConnectionLostOverlay` component (its own 5-attempt exponential-backoff reconnection loop against `GET /users/me`, Retry + Reload buttons) from `DesktopApp.tsx`.
- [x] Removed: the pre-existing hard-outage state machine in `dbHealthMonitor`, the `dbHealthy` / `confirmedUnhealthy` flags, the 10s `consecutiveErrorTimer` that flipped the client to "unhealthy" on sustained failures, and the `database-connection-lost` / `database-connection-restored` events that drove the overlay. The monitor now only emits `database-connection-degraded` / `database-connection-degraded-cleared`.
- [x] Removed: the `dbConnectionFailed` local state in `DesktopApp.tsx` that gated on the hard-outage events, and the `authLoading && !dbConnectionFailed` short-circuit guarded by it.
- [x] Fixed: `POST /activity/log` returned 500 on every call for users with more than 100 recorded activities, the trim-over-100 path called `SimpleDBOps.delete(recentActivity, "recent_activity", userId)`, passing the userId as the `where` argument. Drizzle then saw an invalid clause and threw an error. Now uses `inArray(recentActivity.id, idsToDelete)` with the actual IDs and runs inside a best-effort try/catch so a trim failure no longer fails the primary log insert.
- [x] Fixed: `POST /metrics/register-viewer` occasional 500s on unresolvable hosts or internal errors (see the two "Updated" items above).
- [x] Fixed: random `GET /status` `ERR_NETWORK` failures that previously threw the entire client into the overlay on the first failed attempt, now covered by the progressive retry and the non-blocking toast UX above.

# Changes Made

- `src/backend/dashboard.ts`: `/activity/log` trim loop rewritten with `inArray` + best-effort try/catch.
- `src/backend/ssh/server-stats.ts`: `/metrics/register-viewer` validates the host and stats config before registering, returns `skipped: true` with a reason when there is nothing to poll. `PollingManager.registerViewer` hardened against unhandled rejections.
- `src/ui/main-axios.ts`: new `STATUS_RETRY_SCHEDULE` + `isTransientStatusError` helper drive `getAllServerStatuses`; a request-scoped `__silentRetry` flag lets intermediate attempts skip the response-interceptor logging and the health-monitor side effects. `getServerMetricsById` treats 404 as `null`.
- `src/lib/db-health-monitor.ts`: reduced to a minimal degraded/cleared emitter; "Request aborted" and related codes added to the transient detection.
- `src/ui/desktop/DesktopApp.tsx`: `ConnectionLostOverlay` and its supporting state removed; degraded handler renders a sonner `toast.loading` with `duration: Infinity`, `dismissible: false`, `closeButton: false`, and an action button that reloads the page.
- `src/ui/desktop/apps/dashboard/Dashboard.tsx`: pre-fetches `getAllServerStatuses()` once per cycle and short-circuits offline hosts (and hosts where `registerMetricsViewer` returned `skipped`).
- `src/locales/en.json`: new `common.connectionDegraded` and `common.reload` strings.

# Related Issues

_None found in Termix-SSH/Support._

# Screenshots / Demos

Screenshot of the connection lost toast

<sup>_Screenshot of the new "connection lost" toast, feat. a `Reload` button._</sup>

Checking steps for behavioral changes:

- Open the dashboard with at least one offline host:
  - No `GET /metrics/:id` 404 for that host in the network tab.
  - No more errors related to `POST /activity/log` or `GET /status` unless the retry loop fails.
- While the app is open, `docker pause termix && sleep 30 && docker unpause termix`: the "Server connection lost, recovering…" toast appears, the UI stays fully interactive (tabs, terminals, dialogs all continue to work), and the toast is replaced by the "Server connection restored" success toast on the first healthy response after unpause.

# Checklist

### DIFF
--- a/src/backend/dashboard.ts
+++ b/src/backend/dashboard.ts
@@ -8,7 +8,7 @@ import {
   hostAccess,
   dashboardPreferences,
 } from "./database/db/schema.js";
-import { eq, and, desc, sql } from "drizzle-orm";
+import { eq, and, desc, sql, inArray } from "drizzle-orm";
 import { dashboardLogger } from "./utils/logger.js";
 import { SimpleDBOps } from "./utils/simple-db-ops.js";
 import { AuthManager } from "./utils/auth-manager.js";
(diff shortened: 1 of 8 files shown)


LukeGus:
LGTM, thank you!

PR #815, opened 2026-05-25
persist and use jwt for embedded mode

pgpbj:
# Overview

- [ ] Fixed: Local server mode on the main branch is completely broken for me.  I cannot add hosts, or do anything really.  These changes essentially get this mode working by adding the jwt to the requests going to the internal server. 

# Changes Made

The change in main-axios.ts stores the jwt coming from the user/login request to local storage so that it can later be included or appended to the requests.  I don't understand how this code is supposed to work, so the change to this file is only speculative -- I assume there is a better fix.  

Once I was able to create a host in local server mode, I immediately started getting "Authentication failed" errors when trying to connect to the host via SSH.  This is why the fix in Terminal.tsx was required.

I tried porting these changes to the code from the dev-2.3.0 branch.  They improved the behavior by allowing me to login and create hosts.  However, it looks like there may be some other issue in that branch as SSH connections seem to hang and timeout.

# Related Issues

# Screenshots / Demos

# Checklist

### DIFF
--- a/src/ui/desktop/apps/features/terminal/Terminal.tsx
+++ b/src/ui/desktop/apps/features/terminal/Terminal.tsx
@@ -851,6 +851,13 @@ const TerminalInner = forwardRef<TerminalHandle, SSHTerminalProps>(
 
         if (isEmbeddedMode()) {
           baseWsUrl = "ws://127.0.0.1:30002";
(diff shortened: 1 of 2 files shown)


LukeGus:
Changes ported to dev-2.3.0 branch. Thanks for the fix!

PR #816, opened 2026-05-26
feat(tabs): rename tabs via double-click

Detrol:
## Summary

Adds inline tab rename on double-click, requested as a Tabby-style ergonomic
addition where each session can carry its own label distinct from the host name.

- Double-click a tab label → input with current title selected
- Enter or blur → save
- Esc → cancel
- Empty input → clear custom title (falls back to host/default name)

A new `customTitle?: string` field on `TabContextTab` holds the user override.
Display logic prefers `customTitle` over the existing `title` (host name or
default). `updateHostConfig` is unchanged so renaming the host no longer
overwrites a custom tab title.

Persistence rides on the existing `termix_tabs` localStorage flow, so renames
survive reloads when terminal session persistence is enabled (default).

## Scope

- Desktop only this iteration. Mobile UI tab rendering lives in
  `BottomNavbar.tsx` and double-click is awkward on touch; happy to add a
  long-press variant in a follow-up if maintainers prefer.
- Currently wired for the terminal/server_stats/file_manager/rdp/vnc/telnet/
  tunnel/docker/user_profile render block. ssh_manager/admin/network_graph
  are practically singletons and were left out to keep this PR small; trivial
  to extend if desired.
- `nav.doubleClickToRename` added to `en.json` only — other locales fall
  through Crowdin as usual.

## Drive-by

While touching `TopNavbar.tsx` I removed a stray `(tab.hostConfig as any)?.enableFileManager`
cast — `SSHHost` already declares `enableFileManager: boolean` so the cast was
unnecessary and was tripping the `check-any-changed` lint rule.

## Test plan

- [x] Double-click terminal tab → edit, Enter saves, Esc cancels, blur saves
- [x] Empty input clears custom title and reverts to host name
- [x] Updating host config (rename in Host Manager) preserves custom title
- [x] Reload with persistence enabled keeps custom title
- [x] Built and ran via `docker/Dockerfile`, exercised against a live SSH host

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -568,7 +568,8 @@
     "sudoPasswordCopied": "Sudo password copied to clipboard",
     "noPasswordAvailable": "No password available",
     "failedToCopyPassword": "Failed to copy password",
-    "openFileManager": "Open File Manager"
+    "openFileManager": "Open File Manager",
+    "doubleClickToRename": "Double-click to rename"
   },
   "admin": {
     "title": "Admin Settings",
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -492,6 +492,7 @@ export interface TabContextTab {
     | "vnc"
     | "telnet";
   title: string;
+  customTitle?: string;
   hostConfig?: SSHHost;
   terminalRef?: RefObject<TerminalRefHandle | null>;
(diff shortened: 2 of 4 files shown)


Detrol:
Closing — no longer needed on my end.

PR #864, opened 2026-05-31
fix: patch TOTP bypass and password hash disclosure

ZacharyZcR:
## Summary
- Remove `passwordHash` from `/users/list` API response — hashes should never leave the server
- Require **both** password and TOTP code for `/totp/disable` and `/totp/backup-codes` — previously either alone was sufficient, defeating MFA

## Security Advisories Addressed
- GHSA-ccm8-q457-6vjj (password hash disclosure)
- GHSA-wqfw-rqj7-fv9m (TOTP bypass via password-only)

## Test plan
- [ ] Verify `/users/list` no longer returns `passwordHash` field
- [ ] Verify `/totp/disable` rejects requests with password only (no TOTP code)
- [ ] Verify `/totp/backup-codes` rejects requests with password only
- [ ] Verify both endpoints still work when both password + valid TOTP code are provided

### DIFF
--- a/src/backend/database/routes/users.ts
+++ b/src/backend/database/routes/users.ts
@@ -3209,8 +3209,10 @@ router.post("/totp/disable", authenticateJWT, async (req, res) => {
   const userId = (req as AuthenticatedRequest).userId;
   const { password, totp_code } = req.body;
 
-  if (!password && !totp_code) {
-    return res.status(400).json({ error: "Password or TOTP code is required" });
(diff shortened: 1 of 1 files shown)


ZacharyZcR:
Both fixes already landed in upstream dev-2.3.2 via route refactor (user-admin-routes.ts removed passwordHash, user-totp-routes.ts uses AND logic). Closing as resolved.

PR #901, opened 2026-06-10
fix: adjust RDP desktop height and implement CJK font fallbacks

hanakokoizumi:
# Overview

Two quality-of-life improvements: locale-aware CJK font rendering for terminals and the UI, plus a fix for RDP desktop height calculations that account for the tab bar and mobile bottom bar.

- [x] Added: Locale-aware CJK font fallback system (src/ui/lib/fonts.ts) with per-locale font stacks (zh-CN, zh-TW, ja, ko)
- [x] Added: Unit tests for CJK font fallback and terminal font resolution
- [x] Added: getGuacamoleDisplaySize() and waitForGuacamoleDisplaySize() helpers for accurate RDP viewport sizing
- [x] Updated: Terminal, TerminalPreview, and ConsoleTerminal components to use resolveTerminalFontFamily() based on current i18n locale
- [x] Updated: --font-mono CSS custom property to include CJK fallbacks via var(--font-cjk)
- [x] Updated: GuacamoleDisplay to observe tab bar and mobile bottom bar for correct desktop height
- [x] Removed: Redundant fallback fields from TERMINAL_FONTS entries (now centralized in fonts.ts)
- [x] Fixed: RDP desktop being clipped or misaligned when the tab bar or mobile bottom bar is visible

# Changes Made

### CJK Font Fallbacks (9bf2bb1)

Previously, terminal font stacks only listed Latin monospace fonts. CJK characters rendered with whatever the OS default was, often using the wrong regional typeface (e.g. simplified Chinese glyphs in a ja-JP terminal).

The new fonts.ts module provides getCjkFontFallback(locale) which returns a prioritized font stack per locale:
- zh-CN / zh — PingFang SC, Microsoft YaHei, Noto Sans SC
- zh-TW / zh-HK — PingFang TC, Microsoft JhengHei, Noto Sans TC
- ja — Hiragino Sans, Yu Gothic UI, Meiryo, Noto Sans JP
- ko — Apple SD Gothic Neo, Malgun Gothic, Noto Sans KR
- fallback — balanced stack of all CJK families

resolveTerminalFontFamily(primaryFont, locale) wraps the primary font with Latin and CJK fallbacks. The font stack is applied reactively via i18n hooks — when the language changes, --font-cjk is updated on :root and all terminals pick up the new fallbacks.

TERMINAL_FONTS entries no longer carry their own fallback strings, removing duplication.

### RDP Desktop Height Fix (2daf851)

The Guacamole RDP display used to measure container.clientHeight directly for sendSize, which didn't account for the tab bar (above) or mobile bottom bar (below). On mobile or with the tab bar open, the remote desktop would appear clipped at the bottom.

New logic in guacamole-display-size.ts:
- getGuacamoleDisplaySize() clamps the container height to the visual viewport bottom and to the mobile bottom bar top, ensuring the remote desktop fits within the actual visible area
- waitForGuacamoleDisplaySize() polls up to 60 frames (replacing a fixed 2-frame requestAnimationFrame wait) for a stable measurement ≥100px² — handles slow layouts more robustly
- ResizeObservers are attached to the tab bar ([data-termix-tab-bar]) and mobile bottom bar ([data-termix-mobile-bottom-bar]) so the display recalculates when either bar appears/disappears
- visualViewport resize and scroll events also trigger resync, critical for mobile keyboards pushing the viewport up

Alignment fix: The display wrapper changed from items-center to items-start so the remote desktop stays top-aligned rather than floating mid-container when the container is taller than the scaled display.

# Related Issues

None

# Screenshots / Demos

### Before:
屏幕截图_20260610_170559
屏幕截图_20260610_171148

### After:
屏幕截图_20260610_171532
屏幕截图_20260610_170140

# Checklist

- [x] Supports mobile and desktop UI/app (RDP fix specifically tested on mobile viewport; CJK fonts tested across all four locales)
- [x] I have read Contributing.md (https://github.com/Termix-SSH/Termix/blob/main/CONTRIBUTING.md)
- [x] This is not a translation request

### DIFF
--- a/src/ui/features/docker/components/ConsoleTerminal.tsx
+++ b/src/ui/features/docker/components/ConsoleTerminal.tsx
@@ -20,7 +20,6 @@ import type { SSHHost } from "@/types";
 import { isElectron } from "@/main-axios.ts";
 import { SimpleLoader } from "@/lib/SimpleLoader.tsx";
 import { useTranslation } from "react-i18next";
-import { resolveTermixThemeColors } from "@/features/terminal/terminal-theme";
 import {
   TERMINAL_THEMES,
   DEFAULT_TERMINAL_CONFIG,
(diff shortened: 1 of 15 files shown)


LukeGus:
The RDP height calculation was already fixed. If you could remove that commit and keep the other fix, that would be great. Then I'll go ahead and review it.

PR #929, opened 2026-06-22
fix: file upload fails with 400 and missing schema migrations on v2.4.1 upgrade

### LINKED ISSUE #928: v2.4.1: File upload fails with 400 'Expected multipart/form-data request'
## Description

File upload via the file manager is broken in v2.4.1. Uploading any file returns a 400 error:

```
POST /ssh/file_manager/ssh/uploadFileStream 400 (Bad Request)
Expected multipart/form-data request
```

## Root Cause

In `src/ui/api/ssh-file-operations-api.ts`, the `uploadSSHFileStream` function was refactored in v2.4.1 from a JSON-based upload to a FormData/streaming upload (for the 100MB upload size support). However, the call to `fileManagerApi.post()` does not override the `Content-Type` header:

```typescript
// v2.4.1 — broken
const response = await fileManagerApi.post("/ssh/uploadFileStream", form, {
    timeout: 0,
});
```

All axios instances in Termix are created with `headers: { "Content-Type": "application/json" }` as the default (in `createApiInstance` in `main-axios.ts`). When axios sees `FormData` + `Content-Type: application/json` default, its `transformRequest` JSON-serializes the FormData instead of sending it as multipart — the file object becomes `{}` and the binary is lost. The backend correctly rejects it.

Compare with other functions in the same file that handle this correctly by explicitly setting the header, e.g. the SSH key upload:

```typescript
// Correct — explicit Content-Type override
await G.post(`/db/host`, t, { headers: { "Content-Type": `multipart/form-data` } })
```

In v2.4.0 the upload used `POST /ssh/uploadFile` with a plain JSON body, which worked fine.

## Fix

Change `fileManagerApi.post` to `fileManagerApi.postForm` — the axios method that sets `Content-Type: multipart/form-data` explicitly:

```typescript
// src/ui/api/ssh-file-operations-api.ts
const response = await fileManagerApi.postForm("/ssh/uploadFileStream", form, {
    timeout: 0,
});
```

`postForm` is provided by axios exactly for this case. One-word change.

## Also in v2.4.1: Missing DB migrations

Two schema items were added to `schema.ts` but not to the `migrateSchema()` function in `db/index.ts`, causing 500 errors on existing installations:

1. `GET /service-links` → `SqliteError: no such table: dashboard_service_links`
2. `GET /user-preferences` → `SqliteError: no such column: "status_color_scheme"`

**Fix** — add to `migrateSchema()` in `src/backend/database/db/index.ts`:

```typescript
addColumnIfNotExists("user_preferences", "status_color_scheme", "TEXT");
```

And create the table on startup if it doesn't exist:

```typescript
db.prepare(`CREATE TABLE IF NOT EXISTS dashboard_service_links (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
  label TEXT NOT NULL,
  url TEXT NOT NULL,
  "order" INTEGER NOT NULL DEFAULT 0,
  created_at TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP
)`).run();
```

## Environment

- Version: `ghcr.io/lukegus/termix:release-2.4.1`
- Reproduced on a fresh Docker install upgrading from v2.4.0
- Previous version (v2.4.0) did not have either issue

neoformalex:
## Summary

Fixes two bugs introduced in v2.4.1 (closes #928).

### Bug 1 — File upload returns 400 `Expected multipart/form-data request`

**Root cause:** `uploadSSHFileStream` in `ssh-file-operations-api.ts` was refactored in v2.4.1 from a JSON-based upload to a `FormData` streaming upload (to support the new 100MB limit). However, all axios instances in Termix are created with `Content-Type: application/json` as the default header. When axios sees `FormData` + `Content-Type: application/json`, its `transformRequest` JSON-serializes the FormData instead of sending it as multipart — the `File` object becomes `{}` in the body and the binary is lost.

**Fix:** Change `fileManagerApi.post()` to `fileManagerApi.postForm()`. The `postForm` method sets `Content-Type: multipart/form-data` so `transformRequest` returns the `FormData` as-is, and the browser XHR automatically assigns the correct `multipart/form-data; boundary=...` header.

```diff
- const response = await fileManagerApi.post("/ssh/uploadFileStream", form, {
+ const response = await fileManagerApi.postForm("/ssh/uploadFileStream", form, {
    timeout: 0,
  });
```

Note: other functions in the same file that use `FormData` (e.g. SSH key upload) correctly pass explicit `Content-Type: multipart/form-data` — this one was missed during the v2.4.1 refactor.

### Bug 2 — 500 errors on `/service-links` and `/user-preferences` after upgrade

**Root cause:** Two new schema items were added to `schema.ts` in v2.4.1 but not to the `migrateSchema()` function in `db/index.ts`. Existing installations upgrading from v2.4.0 hit:
- `SqliteError: no such table: dashboard_service_links`
- `SqliteError: no such column: "status_color_scheme"`

**Fix:** Add the missing entries to `migrateSchema()`:

```diff
  addColumnIfNotExists("user_preferences", "compact_host_view", "INTEGER");
+ addColumnIfNotExists("user_preferences", "status_color_scheme", "TEXT");
+
+ sqlite.exec(`
+   CREATE TABLE IF NOT EXISTS dashboard_service_links (
+     id INTEGER PRIMARY KEY AUTOINCREMENT,
+     user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+     label TEXT NOT NULL,
+     url TEXT NOT NULL,
+     "order" INTEGER NOT NULL DEFAULT 0,
+     created_at TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP
+   )
+ `);
```

## Test plan

- [ ] Upgrade a v2.4.0 install to v2.4.1 — `/service-links` and `/user-preferences` should no longer return 500
- [ ] Upload a file via the file manager — should succeed instead of returning 400
- [ ] Fresh v2.4.1 install — both endpoints and upload work normally

### DIFF
--- a/src/backend/database/db/index.ts
+++ b/src/backend/database/db/index.ts
@@ -695,6 +695,18 @@ const migrateSchema = () => {
   addColumnIfNotExists("user_preferences", "confirm_tab_close", "INTEGER");
   addColumnIfNotExists("user_preferences", "hidden_rail_tabs", "TEXT");
   addColumnIfNotExists("user_preferences", "compact_host_view", "INTEGER");
+  addColumnIfNotExists("user_preferences", "status_color_scheme", "TEXT");
+
+  sqlite.exec(`
+    CREATE TABLE IF NOT EXISTS dashboard_service_links (
+      id INTEGER PRIMARY KEY AUTOINCREMENT,
+      user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+      label TEXT NOT NULL,
+      url TEXT NOT NULL,
+      "order" INTEGER NOT NULL DEFAULT 0,
+      created_at TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP
+    )
+  `);
 
   addColumnIfNotExists("users", "is_admin", "INTEGER NOT NULL DEFAULT 0");
 
--- a/src/ui/api/ssh-file-operations-api.ts
+++ b/src/ui/api/ssh-file-operations-api.ts
@@ -357,7 +357,7 @@ export async function uploadSSHFile(
     if (userId !== undefined) form.append("userId", userId);
     form.append("file", file, fileName);
 
-    const response = await fileManagerApi.post("/ssh/uploadFileStream", form, {
+    const response = await fileManagerApi.postForm("/ssh/uploadFileStream", form, {
       timeout: 0,
     });
     return response.data;


LukeGus:
LGTM, thanks.

PR #931, opened 2026-06-23
fix: migrate status color preference column

ZacharyZcR:
## Summary
- add the missing `status_color_scheme` column to new `user_preferences` tables
- migrate existing databases so user preference reads/writes no longer fail after upgrading

Refs https://github.com/Termix-SSH/Support/issues/871

## Validation
- npm run type-check
- npx eslint src/backend/database/db/index.ts src/backend/database/routes/user-preferences.ts
- git diff --check


LukeGus:
Already been fixed.

PR #945, opened 2026-06-23
Guard Guacamole disconnect during startup

ZacharyZcR:
## Summary
- Make Guacamole disconnect idempotent and catch disconnect failures during tab teardown
- Stop async startup work from mutating state after the display unmounts
- Ignore Guacamole state/error callbacks after close so closing a connecting VNC tab cannot blank the whole UI

Fixes Termix-SSH/Support#856

## Verification
- npx prettier --check src/ui/features/guacamole/GuacamoleDisplay.tsx
- npx eslint src/ui/features/guacamole/GuacamoleDisplay.tsx
- npm run type-check
- git diff --check

## Build note
- npm run build still fails before this change completes because src/ui/index.css cannot resolve @fontsource/jetbrains-mono/400.css in the current checkout.


LukeGus:
@copilot resolve the merge conflicts in this pull request

PR #948, opened 2026-06-26
fix: chunked upload for files >=1.5GB to bypass browser ArrayBuffer limit

### LINKED ISSUE #870: [BUG]
### Title

Web file manager fails to upload files ≥2GB — DOMException: File could not be read

### Platform

Website - Firefox

Docker

### Version

2.4.0

### Troubleshooting

- [ ] I have reviewed opened and closed issues
- [ ] I have checked open issues and ensured this is not a duplicate

### The Problem

Uploading a file of ~2GB or larger through the web file manager fails immediately with Upload failed: DOMException: File could not be read, thrown in FileManager-*.js. Files below ~2GB upload fine.
Direct SFTP to the same host transfers the identical 2GB file without issue, so the SSH/SFTP server, destination filesystem, and network path are all healthy. The failure is isolated to the browser-side upload path.
Also reproduces when connecting directly to the backend on :8080, bypassing my reverse proxy (NPMplus) entirely — so the proxy is not involved.
Reproduces in both Firefox and the desktop app (Electron/Chromium), i.e. both browser engines — pointing at the shared web upload code rather than a browser-specific issue.
Likely cause (speculative): the 2GB boundary plus the FileReader exception suggest the upload reads the entire file into a single ArrayBuffer before sending, hitting the JS engine's ~2³¹-byte max length. Switching to a streamed/chunked upload (streaming the File/Blob to fetch, or multipart chunks) would lift the limit.

### How to Reproduce

1. Connect to a Linux host with File Manager enabled.
2. Upload a file just under 2GB (e.g. truncate -s 1900M test.bin) → succeeds.
3. Upload a file ≥2GB (truncate -s 2G test.bin) → fails instantly with DOMException: File could not be read.

### Additional Context

Confirmed direct sftp put of the same 2GB file to the same host succeeds, isolating the fault to the web client's upload code. Client: Termix-Desktop/2.3.1 (Linux; Electron/42.2.0). Reproduced both via NPMplus reverse proxy and directly against :8080.

AmSach:
Fixes Termix-SSH/Support#870

## Problem

Uploading a file of ~2GB or larger through the web file manager fails immediately with:

```
Upload failed: DOMException: File could not be read
```

Reproduces in **both** Firefox and the Termix desktop app (Electron/Chromium), so the fault is in the shared browser-side upload code, not a browser-specific quirk. Direct SFTP from the same client to the same host transfers the identical 2GB file without issue, isolating the failure to the web upload path.

The underlying cause is the XHR/ArrayBuffer 2^31 byte ceiling: once the multipart body approaches ~2.14GB the browser refuses to read the underlying File/Blob and throws `DOMException: File could not be read`.

## Fix

I split the upload into two paths in `src/ui/api/ssh-file-operations-api.ts`:

- Files **<= 1.5 GiB**: keep the existing single-pass `/ssh/uploadFileStream` call (unchanged behaviour, no regression for the common case).
- Files **> 1.5 GiB**: slice with `file.slice()` into 8 MiB chunks and POST each one to a new backend endpoint `/ssh/uploadFileChunk` sequentially. `Blob.slice` returns a sub-Blob reference without copying bytes into memory, so each request fits comfortably under the browser ceiling.

The frontend never materializes the whole file in JS memory — we only ever hold one 8 MiB Blob at a time.

## Backend changes

New POST endpoint `/ssh/file_manager/ssh/uploadFileChunk` in `src/backend/ssh/file-manager-content-routes.ts`:

- Multipart form fields: `sessionId`, `path`, `fileName`, `chunkIndex`, `totalChunks`, `totalSize`, `chunk` (file).
- `chunkIndex === 0` opens the remote file with `flags: 'w'` (create/truncate).
- `chunkIndex > 0` opens with `flags: 'a'` (append).
- Streams the chunk directly from busboy into `sftp.createWriteStream()` — no buffering on the server either, so the backend can accept arbitrarily large files.
- Per-part size cap of 32 MiB on the busboy parser as a defensive limit (current client sends 8 MiB chunks).
- Final chunk response includes `complete: true` for clients that want to detect end-of-upload without comparing indices.
- Full OpenAPI doc + structured fileLogger events (`file_upload_chunked_start`, `file_upload_chunk_start`, `file_upload_chunk_complete`, `file_upload_chunked_complete`) for ops visibility.

The existing `/ssh/uploadFileStream` route is **untouched** — clients that hit it directly keep working.

## UX improvement

`handleUploadFile` in `FileManager.tsx` now passes an `onChunkProgress` callback that updates the existing upload toast with per-chunk percentage (`Uploading foo.bin — 47% (chunk 120/256)`) so users get feedback for multi-minute uploads instead of a stuck spinner.

## Tested by

1. `npx tsc --noEmit -p tsconfig.app.json` — no new type errors introduced (pre-existing unrelated errors in this file remain unchanged).
2. `npm run build:backend` — compiles cleanly.
3. `npx vite build` — frontend builds clean, no regressions.
4. `npx eslint src/backend/ssh/file-manager-content-routes.ts src/ui/api/ssh-file-operations-api.ts` — 0 errors, 0 warnings.
5. Wrote a small smoke test for the chunk math: a 2 GiB file yields 256 chunks of 8 MiB with the last chunk exactly 8 MiB and total bytes = 2147483648. A 1.5 GiB file (just at threshold) yields 205 chunks covering 1717986918 bytes. A 1 MiB file stays on the single-pass path.
6. Auth/ownership/validation flow on the new backend route reuses the same `verifySessionOwnership` check as the existing `uploadFileStream` route.

I do not have a Windows/WSL + Firefox test rig handy to do a true end-to-end 2GB upload from the browser, so I'd appreciate a smoke test from the reporter on v2.4.x before this lands.

Fixes Termix-SSH/Support#870

### DIFF
--- a/src/backend/ssh/file-manager-content-routes.ts
+++ b/src/backend/ssh/file-manager-content-routes.ts
@@ -1507,4 +1507,281 @@ export function registerFileContentRoutes(
       req.pipe(bb);
     },
   );
+
+  /**
+   * @openapi
+   * /ssh/file_manager/ssh/uploadFileChunk:
+   *   post:
+   *     summary: Upload a single chunk of a large file
+   *     description: Receives one chunk of a large file upload via multipart form data and either creates a new remote file (chunkIndex=0) or appends to the existing file (chunkIndex>0). Used by the client to upload files larger than ~2GB that would otherwise hit browser-side ArrayBuffer limits.
+   *     tags:
+   *       - File Manager
+   *     requestBody:
+   *       required: true
+   *       content:
+   *         multipart/form-data:
+   *           schema:
+   *             type: object
+   *             required:
+   *               - sessionId
+   *               - path
+   *               - fileName
+   *               - chunkIndex
+   *               - totalChunks
+   *               - totalSize
+   *               - chunk
+   *             properties:
+   *               sessionId:
+   *                 type: string
+   *               path:
+   *                 type: string
+   *               fileName:
+   *                 type: string
+   *               chunkIndex:
+   *                 type: integer
+   *               totalChunks:
+   *                 type: integer
+   *               totalSize:
+   *                 type: integer
+   *               chunk:
+   *                 type: string
+   *                 format: binary
+   *     responses:
+   *       200:
+   *         description: Chunk accepted. When chunkIndex === totalChunks - 1 the full file is complete.
+   *       400:
+   *         description: Missing required parameters, invalid chunk metadata, or SSH connection not established.
+   *       403:
+   *         description: Session access denied.
+   *       500:
+   *         description: Failed to write chunk to remote filesystem.
+   */
+  app.post(
+    "/ssh/file_manager/ssh/uploadFileChunk",
+    (req: Request, res: Response) => {
+      const userId = (req as AuthenticatedRequest).userId;
+
+      const contentType = req.headers["content-type"] || "";
+      if (!contentType.includes("multipart/form-data")) {
+        return res
+          .status(400)
+          .json({ error: "Expected multipart/form-data request" });
+      }
+
+      let sessionId: string | undefined;
+      let remotePath: string | undefined;
+      let fileName: string | undefined;
+      let chunkIndex = -1;
+      let totalChunks = 0;
+      let totalSize = 0;
+      let resolved = false;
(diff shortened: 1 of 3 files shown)


LukeGus:
LGTM, thanks

PR #955, opened 2026-06-26
Fix SSH heading syntax highlighting

ZacharyZcR:
## Summary
- protect bracketed SSH headings like [user@host] from terminal syntax highlighting
- keep log-level highlighting active for text after the heading
- add regression tests for bracketed headings containing log-level words

## Test
- npx vitest run src/ui/lib/terminal-syntax-highlighter.test.ts
- npm run type-check

Fixes Termix-SSH/Support#896


LukeGus:
LGTM, thanks

PR #956, opened 2026-06-26
Fix Proxmox import auth fallback

ZacharyZcR:
## Summary
- avoid importing Proxmox guests as password/key auth when no secret is available
- keep credential auth when a default or discovered credential exists
- add regression coverage for Proxmox import auth resolution

## Test
- npx vitest run src/ui/components/proxmox/proxmox-import-auth.test.ts
- npm run type-check

Fixes Termix-SSH/Support#898


Dimen7 (approved):

LukeGus:
LGTM, thanks

PR #957, opened 2026-06-26
Initialize auth before fullscreen terminal routes

ZacharyZcR:
## Summary
- gate fullscreen feature routes until the app API/session bootstrap completes
- restore the Electron JWT into localStorage before rendering terminal views
- preserve the existing fullscreen login fallback when authentication is unavailable

## Test
- npm run type-check

Fixes Termix-SSH/Support#899


LukeGus:
LGTM, thanks

PR #960, opened 2026-06-26
feat: support SSH agent auth across SSH features

ZacharyZcR:
## Summary
- reuse the existing SSH agent socket resolver to attach ssh2 agents consistently
- support agent auth in file manager, Docker, metrics, tmux monitor, snippets, Proxmox discovery, credential deploy, and jump-host chains
- handle serialized terminalConfig values and cover agent config attachment in tests

## Scope
This addresses the SSH-agent based path for Termix-SSH/Support#821, including YubiKey keys exposed through ssh-agent, gpg-agent, or compatible agent sockets. Direct PKCS#11 library loading is intentionally left out for a separate follow-up.

## Tests
- npm run type-check
- npm run test -- src/backend/ssh/terminal-agent-auth.test.ts
- npx eslint <touched files>

Note: full npm run lint still fails on an existing unrelated unused import in src/backend/ssh/managers/wireguard.ts.


LukeGus:
LGTM, thanks

PR #963, opened 2026-06-26
fix(tunnel): skip endpoint credential validation for direct tunnels

WardenPro:
## Problem

A user connecting to their server **using SSH key authentication only** (no password) sets up a LOCAL tunnel with **"This host (direct tunnel)"** as the endpoint.

When trying to start the tunnel, Termix throws this error:

> Cannot connect tunnel '...': endpoint host requires password authentication but no plaintext password available.

The user never configured a password — the server doesn't have one. They use an SSH key stored in Termix's credential system. The SSH terminal works perfectly fine with that same host, but the tunnel fails every time.

## Root Cause

When building the tunnel connect request, the frontend needs to fill the `endpointAuthMethod` field. For a direct tunnel ("This host"), there is no separate endpoint host to look up, so `endpointSsh` is `undefined`. The code falls back to a hardcoded default:

```js
endpointAuthMethod: endpointSsh?.authType ?? "password"
```

This sends `"password"` to the backend even when the host uses SSH key auth. The backend sees `authMethod: "password"` with no password available and blocks the connection.

## Fix

Two places corrected:

1. **Frontend** — for a direct tunnel, the endpoint is the same host as the source, so use its actual `authType` instead of the `"password"` fallback.
2. **Backend** — defensive guard: skip endpoint credential validation for direct (single-host) tunnels, which never open a second SSH connection.

## Test plan

- [ ] Host configured with SSH key auth (credential or direct key)
- [ ] LOCAL tunnel with "This host (direct tunnel)" → connects without error
- [ ] Tunnel to a separate SSH endpoint → credential validation still works correctly


LukeGus:
LGTM, thanks

PR #964, opened 2026-06-26
fix: use correct translation key for nav.close

reppiz:
# Overview

Fixed invalid translation key for close tab. `nav.closeTab` doesn't exist in the JSON locale files. There is a `nav.close` so I imagine this what we want. 

Using `nav.closeTab` shows up as text when right clicking the existing connection tabs. 

# Changes Made

`/src/ui/shell/TabBar.tsx`

changed line 565 from `{t("nav.closeTab")}` to `{t("nav.close")}`.


LukeGus:
LGTM, thanks

PR #968, opened 2026-06-27
Fix VNC required argument handshake

ZacharyZcR:
## Summary
- handle guacd required/require argument requests in the guacamole-lite patch
- send stored connection settings back through argv/blob/end streams instead of forwarding the request to the browser
- add a regression test that verifies the patched guacamole-lite client handles dynamic argument requests

## Verification
- node scripts/patch-guacamole-lite.cjs
- npm exec vitest run -- scripts/patch-guacamole-lite.test.ts
- npx prettier --check .
- npm run lint
- npx tsc --noEmit
- npm run build

Fixes Termix-SSH/Support#906


LukeGus:
LGTM, thanks

PR #969, opened 2026-06-27
Prioritize host results in command palette search

ZacharyZcR:
## Summary
- move the Servers and Hosts results block to the top of the command palette while a search query is active
- keep the existing Quick Actions / Recent Activity first layout when the search field is empty

## Verification
- npx prettier --check .
- npx tsc --noEmit
- npm run lint
- npm run build

Fixes Termix-SSH/Support#902


LukeGus:
LGTM, thanks

PR #970, opened 2026-06-27
Prevent sidebar host hover layout shift

ZacharyZcR:
## Summary
- make sidebar host hover action trays float outside the row layout
- keep click/touch action trays expanding as before
- apply the same no-layout-shift behavior to compact and regular host rows

## Verification
- npx prettier --check .
- npx tsc --noEmit
- npm run lint
- npm run build

Fixes Termix-SSH/Support#903


LukeGus:
LGTM, thanks

PR #973, opened 2026-06-27
Make file downloads reliable in desktop app

ZacharyZcR:
## Summary
- keep blob download URLs alive briefly after triggering a file-manager download
- append the temporary download anchor to the document before clicking it for better Electron/Windows reliability
- leave the backend download stream unchanged

## Notes
- The upload side of Support#901 is already covered by #929/#932. This PR addresses the remaining fragile client-side download trigger path.

## Verification
- npx prettier --check .
- npx tsc --noEmit
- npx eslint src/ui/api/ssh-file-operations-api.ts
- npm run lint
- npm run build

Refs Termix-SSH/Support#901


LukeGus:
LGTM, thanks

PR #980, opened 2026-06-28
Fix shared snippet folder visibility

### LINKED ISSUE #879: [BUG] Snippet Sharing
### Platform

TrueNas Scale Apps - Ubuntu VM

### Version

2.4.1_1.2.12

### The Problem

Sharing Snippets that are already in a folder does not show until the user creates a folder with the exact same name - Possibly look to share the folder and all snippets in that folder ?

### How to Reproduce

Sharer
1.Snippets
2.Open Folder 
3.Share Snippet
4.User or Role Share
User/Roles
1.Snippets
2.Dont see the shared snippet
3.Create Folder - Exact Same Folder Name as the Sharers folder name
4.Snippet is then seen in the folder

---

> [!NOTE]
> Submitted via the Termix Discord Support Bot
> Reported by snicher_za on Discord
> [View Discord thread](https://discord.com/channels/1347374268253470720/1519001038219120811)

kavix:
Resolves Termix-SSH/Support#879 by introducing virtual folders in the sidebar to prevent shared snippets from disappearing if their folder doesn't exist for the recipient.

### DIFF
--- a/src/types/ui-types.ts
+++ b/src/types/ui-types.ts
@@ -223,11 +223,7 @@ export type Tab = {
 };
 
 export type DockerContainerStatus =
-  | "running"
-  | "exited"
-  | "paused"
-  | "created"
-  | "restarting";
+  "running" | "exited" | "paused" | "created" | "restarting";
 
 export type DockerContainer = {
   id: string;
@@ -276,10 +272,7 @@ export type LayoutPreset = {
 };
 
 export type UserProfileSection =
-  | "account"
-  | "appearance"
-  | "security"
-  | "api-keys";
+  "account" | "appearance" | "security" | "api-keys";
 export type AdminSection =
   | "general"
   | "sso"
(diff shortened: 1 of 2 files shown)


ZacharyZcR (changes requested):
Thanks for the contribution. This is close, but I don't think it is merge-ready yet.

Blocking issues:
- Shared snippets whose folder does not exist locally are now included in  and also rendered again under the new virtual folder. That makes the same shared snippet appear twice.  should stay limited to snippets with no folder, while virtual folders should handle snippets with a non-empty missing folder.
- Checking formatting... fails on , so CI would fail once checks run.

Also, please keep the PR scoped to the missing-folder visibility bug. Virtual folders should not be editable/deletable, and they should not become selectable as real destination folders unless that behavior is intentional.

ZacharyZcR:
Clarifying the review above because the code identifiers were stripped from the first message:

- `uncategorizedSnippets` currently includes snippets whose folder name is missing locally, while the same snippets are also rendered under the virtual folder. That duplicates shared snippets in the sidebar. `uncategorizedSnippets` should only include snippets with no folder; snippets with a non-empty missing folder should be shown only under their virtual folder.
- `npx prettier --check src/ui/sidebar/SnippetsPanel.tsx src/types/ui-types.ts` currently fails on `src/ui/sidebar/SnippetsPanel.tsx`.
- Please keep the change scoped to visibility. Virtual folders should not be editable/deletable, and they should not be selectable as real destination folders unless that behavior is intentional.

ZacharyZcR:
Thanks for the update. The duplicate display issue looks fixed now.

There are still two blockers before this can be merged:

- The PR is currently closed, so it cannot be merged unless it is reopened or recreated.
- `npx prettier --check src/ui/sidebar/SnippetsPanel.tsx src/types/ui-types.ts` still fails, now on `src/types/ui-types.ts`.

Please also remove unrelated changes from `src/types/ui-types.ts` and keep the PR scoped to the shared snippet folder visibility fix.

PR #983, opened 2026-06-28
Fix TUI terminal rendering with syntax highlighting

ZacharyZcR:
## Summary
- skip terminal syntax highlighting while alternate-screen TUI apps are active
- bypass highlighting for chunks containing private TUI control sequences such as alternate screen, cursor visibility, and bracketed paste modes
- add highlighter coverage for private TUI escape sequences

## Testing
- npm run test -- src/ui/lib/terminal-syntax-highlighter.test.ts
- npm run type-check
- npm run lint
- npm run format:check
- npm run biome:check
- npm run build

Fixes Termix-SSH/Support#910


Dark-Sarmat:
First of all, thank you for the quick fix and for the 2.5.0 release!

I tested it immediately after upgrading.

The original issue is mostly fixed. Midnight Commander is now usable again.

Current status:

- ✅ Version 2.5.0 no longer corrupts the screen immediately.
- ✅ `mc -a` works perfectly.
- ✅ Normal `mc` also works correctly when **Syntax Highlighting** is disabled.
- ❌ With **Syntax Highlighting** enabled, MC still develops rendering artifacts during normal navigation.

The corruption is now much more subtle than before. It appears gradually while browsing directories or refreshing the panels, instead of breaking immediately.

So the remaining issue seems to be related specifically to the Syntax Highlighting parser interacting with MC's screen redraws.

For now I can work around the problem by disabling Syntax Highlighting, but I would really like to keep using this feature because it is one of the nicest parts of Termix.

Thank you again for the very fast response and for fixing the biggest part of this issue.

PR #984, opened 2026-06-28
Rework SSH credential password handling

ZacharyZcR:
## Summary
- allow credentials and key-based hosts to store an optional account password alongside SSH key details
- retry SSH terminal connections once with the stored password when key authentication fails
- replace the credential auth type toggle with one auth page for password and/or key details
- remove the host terminal Sudo Password Auto-Fill setting and force saved host payloads to keep silent auto-fill disabled
- add a manual Fill password action in SSH Tools and a confirm-to-fill prompt for terminal password prompts

## Testing
- npm run test -- src/ui/sidebar/HostEditorData.test.ts
- npm run type-check
- npm run lint
- npm run format:check
- npm run biome:check
- npm run build

Fixes Termix-SSH/Support#593


JHarding86:
Running version 2.5.0 Stable

So this feature works when setting up a credential and then using that credential on a host, but is does not work when just setting up an SSH key and password within a host. No matter what I do, the optional password does not get saved when I update the host. I presume that because the password does not save, that when I try to autofill, I get the "No saved passwords" toast.

image


image

Issue #1063, opened 2026-07-15
API keys cannot access encrypted data (credentials/hosts) — "User data not unlocked" for headless automation


Kalvalax:
## Summary

API keys (v2.3+, advertised as "user-scoped API keys … to be used for
automation/CI") authenticate successfully but **cannot perform any operation
that touches an encrypted field** — creating/reading SSH credentials or hosts
fails with HTTP 500 `User <id> data not unlocked`, unless that user *also*
happens to have a live interactive (password/OIDC) session that already
unlocked their Data Encryption Key (DEK) in server memory.

This makes API keys unusable for true headless automation against the
credential/host store, which is the main thing automation would want to do.

## Environment

- Termix v2.3 through **v2.5.0** (verified still reproducible in `release-2.5.0`, commit `9de904d`; `ghcr.io/lukegus/termix:latest`). In 2.5.0 `handleApiKeyAuth` moved to ~L756 of `auth-manager.ts` but the DEK gap is unchanged, and `deriveOIDCSystemKey`/`authenticateOIDCUser` are byte-identical.
- Deployment: multi-user with **OIDC (Authentik)** as the identity provider
- Use case: an external provisioning system creating/updating SSH host entries
  + credentials on behalf of users, headless (no browser session).

## Root cause (traced through the source)

The per-user DEK is only loaded into `userSessions` on an **interactive login**
(`authenticateUser` / `authenticateOIDCUser`). The API-key auth path does not
load it:

`src/backend/utils/auth-manager.ts` — `handleApiKeyAuth(...)` ends with:

```ts
req.userId = matchedKey.userId;
next();
```

No call to `authenticateOIDCUser` / `setupOIDCUserEncryption` /
`restoreUserDataKey`, so the user's DEK session is never established.

Every encrypted-table write then hits the hard gate in
`src/backend/utils/simple-db-ops.ts` → `SimpleDBOps.insert(...)`, whose first
line is:

```ts
const userDataKey = DataCrypto.validateUserAccess(userId);
```

and `src/backend/utils/data-crypto.ts`:

```ts
static validateUserAccess(userId: string): Buffer {
  const userDataKey = this.getUserDataKey(userId);
  if (!userDataKey) {
    throw new Error(`User ${userId} data not unlocked`);
  }
  return userDataKey;
}
```

`ssh_credentials` (`password`, `privateKey`, `key`, `keyPassword`, `publicKey`)
and `ssh_data` (host secrets) are encrypted tables
[...]

PR #1064, opened 2026-07-15
feat(auth): opt-in OIDC DEK unlock for API-key requests (ALLOW_APIKEY_DATA_UNLOCK)

Kalvalax:
Fixes the gap in #1063: API keys authenticate but cannot access the encrypted credential/host store (`User data not unlocked`), so they're unusable for headless automation unless the user also has a live interactive session.

### What this does
For **OIDC-backed users**, `handleApiKeyAuth` now unlocks the per-user Data Encryption Key after matching the key, using the server-derivable system key (`deriveOIDCSystemKey`) — no user password needed. Encrypted-table operations (credentials/hosts) then work for API-key requests.

### Opt-in, OFF by default
Gated behind **`ALLOW_APIKEY_DATA_UNLOCK`** (default `false`). Enabling it widens the blast radius of a leaked API key — it can then read that user's stored SSH secrets in clear text — so it must be turned on explicitly by operators who run automation. Default behaviour is unchanged (fail-closed).

- **OIDC-only:** password users are deliberately not unlocked (their KEK is password-derived and cannot be reproduced server-side).
- Mirrors the existing `requireAdmin` DB lookup pattern in the same method; failures are logged and non-fatal (the request proceeds unlocked, i.e. the old behaviour).

### Notes / open for discussion
- A short unlock TTL / per-request unlock could further limit exposure (currently reuses `authenticateOIDCUser`'s session).
- Happy to adjust the env var name / add docs if you prefer.

Context: we run Termix headless behind an OIDC IdP (Authentik) with an external provisioning system creating SSH hosts/credentials on behalf of users; we've been carrying this as a local patch. See #1063 for the full root-cause trace.


LukeGus:
LGTM, thanks.

PR #1074, opened 2026-07-21
Fix Guacamole tab visibility lifecycle

default-student:
_Disclaimer: The investigation and fix was conducted using codex gpt-5.5 medium. No human code here. However the fix does work for me and makes VNC usable. Initially I intended to keep it private, but I hope this is at least helpful to others. Thank you for your software._

## Summary

This fixes several Guacamole/VNC connection paths that could leave Termix stuck on `Connecting...` with no rendered remote desktop.

VNC initially did not render at all for the reproduced host, even though the backend and guacd appeared to be doing work. After fixing the first blockers, close/reopen, reconnect, and tab switching exposed additional lifecycle bugs in the same Guacamole frontend path.

The fix does three things:

- makes the VNC/RDP frontend connect path more robust in Electron
- makes embedded Guacamole tabs respect the real active-tab visibility state
- removes VNC options/recording behavior that caused guacd/x11vnc sessions to close during startup

## What Was Happening

We reproduced this with a real VNC host behind Termix. The backend API returned a token, guacd accepted the connection, and x11vnc showed a client briefly connecting, but the Termix UI stayed on `Connecting to VNC session...` instead of rendering the desktop.

This was not a single backend outage or a missing guacd service. During debugging we found several related failure modes in the VNC/Guacamole path:

1. VNC could fail on first open: guacd and x11vnc saw connection activity, but the frontend stayed on the connecting loader.
2. The websocket URL was built with token/size query params, then `client.connect()` was called without data. `guacamole-common-js` appends the connect data to the tunnel URL, so this could produce malformed URLs ending in `?undefined`.
3. The initial layout wait and retry loop depended on `requestAnimationFrame()`. In Electron, especially around inactive windows/tabs, those callbacks can be delayed or skipped long enough that no Guacamole display is attached.
4. `guacamole-common-js` gates display frame flushing on `document.hasFocus()`. In Electron this can leave received display frames waiting behind a frame callback while guacd continues sending data.
5. If guacd disconnected before the client reached `CONNECTED`, `GuacamoleDisplay` cleared the ready state but left the internal connecting guard set. That made reconnects look like they were running while the actual connect path was blocked.
6. Hidden Guacamole tabs were treated as visible because `tabUtils` did not pass `isVisible` into `GuacamoleApp`; `GuacamoleApp` then always passed `isVisible={true}` to `GuacamoleDisplay`.
7. After the first-open issues were fixed, switching away from VNC and back exposed that hidden-tab lifecycle bug: the VNC tab could reconnect or resize while inactive and return to `Connecting...`.
8. VNC connections were receiving RDP-only `security: "any"` settings, and automatic Guacamole recording for VNC could make guacd log recording-path errors during startup.

## How This Was Tracked Down

The issue was traced with a local Termix Electron build using Chromium DevTools Protocol plus backend, guacd, and x11vnc logs.

The key observations were:

- Termix created Guacamole tokens successfully.
- guacd reported clients joining and sometimes immediately leaving.
- x11vnc accepted connections but saw the client disappear during protocol startup.
- The renderer sometimes had hidden zero-size canvases while the UI still displayed `Connecting...`.
- Manual websocket/renderer checks showed the backend path could receive Guacamole traffic while the visible UI never became ready.
- Removing invalid VNC backend settings got the remote side further into startup.
- Fixing frontend URL/query handling and Electron frame/layout waits allowed the VNC desktop to render.
- Once VNC worked initially, close/reopen, reconnect, and switching to other tabs exposed additional stale connecting-state and visibility lifecycle bugs.
- Patching the installed app locally confirmed each fix before translating it back into source changes for this PR.

### WebSocket query handling

`GuacamoleDisplay` now passes the base tunnel URL to `new Guacamole.WebSocketTunnel(...)` and passes the token/size query string to `client.connect(query)`.

That matches the `guacamole-common-js` API and avoids malformed websocket URLs where an already-query-bearing URL could get `?undefined` appended by the library.

### Electron startup/layout robustness

The initial display measurement wait no longer depends on nested `requestAnimationFrame()` calls. The short layout waits now use timers, which are more reliable when an Electron window or tab is inactive.

A watchdog also retries the connection if the same Guacamole client is still stuck in the connecting state after 8 seconds. This recovers the early-close case where guacd accepted the client but the frontend never reached a usable display state.

### Disconnect/error state cleanup

When Guacamole reports `DISCONNECTED`, the display now clears the internal connecting guard, marks the display as failed, and surfaces the existing reconnect UI instead of leaving the tab on an endless loader.

### Tab visibility lifecycle

`tabUtils` now passes the actual `isVisible` value to RDP/VNC/Telnet `GuacamoleApp` instances, and `GuacamoleApp` passes it through to `GuacamoleDisplay`.

This lets hidden tabs stop capturing input and avoid visible-tab resize behavior, while keeping the existing Guacamole session alive across tab switches. The tab is not disconnected just because it becomes inactive.

### Dependency display flush patch

The display-frame scheduler that caused one of the Electron-specific hangs lives inside `guacamole-common-js`, not Termix source code. There is no public API to select synchronous flushing.

This PR adds a `postinstall` patch for `guacamole-common-js`, following the same pattern Termix already uses for `guacamole-lite`, `better-sqlite3`, `nan`, and xterm patches. The patch makes Guacamole display flushing synchronous so frame rendering does not depend on `requestAnimationFrame()` being delivered while Electron focus state is unreliable.

### VNC backend options

VNC token/default settings no longer include RDP-only `security: "any"`.

Automatic Guacamole recording is also disabled for VNC. The reproduced VNC setup hit guacd recording-path failures during startup, and VNC should not be forced through recording unless the VNC recording path is known to be valid.

## Testing

- `npm ci` with Node 24/npm 11 via `npx`
- confirmed `scripts/patch-guacamole-common-js.cjs` applies during `postinstall`
- `tsc --noEmit -p tsconfig.json` under Node 24
- `git diff --check HEAD~1..HEAD`
- manually verified the equivalent local Termix patch against a real VNC session:
  - opened VNC successfully after it previously stayed on `Connecting...`
  - closed/reopened VNC in the same Electron process
  - used reconnect after failed/stuck states
  - switched from VNC to another tab and back
  - verified the VNC canvas returned rendered instead of staying on `Connecting...`

## Notes

A full Vitest run was attempted after installing dependencies. The first run was invalid because it was launched from `/tmp`, and existing tests rely on repo-relative `node_modules/...` fixture paths. It reported 131 passing files before failing on cwd-sensitive fixture lookup. I did not rerun it after that because this PR is currently draft and the rerun was stopped.


default-student:
I assume main is probably the wrong target branch

LukeGus:
LGTM, thanks.

default-student:
🎉

PR #1079, opened 2026-07-25
fix(net): correct SSRF blocklist false-positive blocking all outbound IPv4

brennanneoh:
Fixes [#1024](https://github.com/Termix-SSH/Support/issues/1024)

## Summary

`safeOutboundFetch` was rejecting every outbound IPv4 destination as private, not just the ones that actually are. I hit this through ntfy notification delivery, but it affects anything that goes through `safeOutboundFetch`.

## Root cause

The IPv6 half of the SSRF blocklist has one entry, `["::ffff:0:0", 96]`, meant to catch someone spoofing a private address as an IPv6 literal (a URL containing `::ffff:127.0.0.1`, say, trying to sneak past the plain-IPv4 checks). Node's `BlockList` doesn't respect the family boundary the way you'd expect: it compares addresses through their IPv4-mapped-IPv6 form no matter what `type` you pass to `check()`. That's documented, just easy to miss ([BlockList docs](https://nodejs.org/api/net.html#class-netblocklist) show the mirror case, an IPv4 rule catching an IPv6-literal check). `::ffff:0:0/96` is the entire IPv4 address space once mapped, so that one entry was blocking every IPv4 address, including ones checked with `type: "ipv4"` directly.

I found this by patching a running container to log the raw addresses being checked. The failure itself, `"Private destinations are not allowed"` on a plain public Cloudflare IP, gave no hint the bug was in the blocklist rather than DNS, SSRF logic, or the request itself.

## Changes

- Replaced the blanket `::ffff:0:0/96` with individual mapped ranges, one per existing IPv4 CIDR, generated from the same list the IPv4 rules come from. The two lists can't drift apart again the way they did here.
- The lookup hook threw the same `"Private destinations are not allowed"` for two different situations: DNS returning zero addresses, and DNS returning an address that's actually private. Only one of those is a privacy decision. Split into separate error messages.
- Pulled the `connect.lookup` hook out into its own function (`createDnsLookupHook`) so it can be tested against a fake resolver directly, instead of only through a real `fetch()`/`Agent` call.

## User impact

Any notification channel or webhook that resolves to a plain IPv4 address should now go through, instead of failing with a "private destination" error that has nothing to do with the actual destination.

## Testing

Added `src/backend/tests/utils/safe-outbound-fetch.test.ts`:

- `isBlockedAddress`: public/private IPv4, public/private IPv6, IPv4-mapped spoofing of private ranges, IPv4-mapped form of public addresses
- `createDnsLookupHook`: allows a public address through, blocks a private one, distinct error on empty DNS result, real DNS errors pass through untouched
- 11/11 passing, `tsc --noEmit` clean

## Notes

Scoped to the blocklist bug only. While chasing the original notification failure I also found the `connect.lookup` hook doesn't handle `lookupOptions.all`: Node's happy-eyeballs (`autoSelectFamily`) expects the full address array back, not a single address, and this hook always returns a single address. That's a separate bug in the same file. I'll send it as its own PR instead of bundling it here.

### DIFF
--- a/src/backend/tests/utils/safe-outbound-fetch.test.ts
+++ b/src/backend/tests/utils/safe-outbound-fetch.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from "vitest";
+import { isBlockedAddress } from "../../utils/safe-outbound-fetch.js";
+
+describe("isBlockedAddress", () => {
+  it("allows public IPv4 addresses", () => {
+    expect(isBlockedAddress("8.8.8.8")).toBe(false);
+    expect(isBlockedAddress("104.21.52.150")).toBe(false);
+  });
+
+  it("blocks private/reserved IPv4 ranges", () => {
+    expect(isBlockedAddress("10.0.0.1")).toBe(true);
+    expect(isBlockedAddress("172.16.0.1")).toBe(true);
+    expect(isBlockedAddress("192.168.1.1")).toBe(true);
+    expect(isBlockedAddress("127.0.0.1")).toBe(true);
+    expect(isBlockedAddress("169.254.1.1")).toBe(true);
+    expect(isBlockedAddress("100.64.0.1")).toBe(true);
+  });
+
+  it("allows public IPv6 addresses", () => {
+    expect(isBlockedAddress("2606:4700:3034::ac43:c88d")).toBe(false);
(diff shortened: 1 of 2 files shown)


LukeGus:
Thank you, I hadn't noticed that I appreciate it. LGTM and merged for v2.6.0.

PR #1082, opened 2026-07-26
Fix export user record handling and 500 error

### LINKED ISSUE #1020: [BUG] DB Backup error: Failed to export user data
### Platform

Website

### Version

2.5.1

### The Problem

IF I try to export the Database, I am getting the following error: "Failed to export user data"

### How to Reproduce

1. go to Settings > Database > Export Database
2. click "Export"

---

> [!NOTE]
> Submitted via the Termix Discord Support Bot
> Reported by jo_strasser on Discord
> [View Discord thread](https://discord.com/channels/1347374268253470720/1530102644721319987)

xyzulu:
This pull request makes small fixes to the user object handling in the `/database/export` API endpoint. The changes ensure that the code consistently uses the `user` object directly rather than treating it as an array.

* Fixed a bug where `user[0]` was incorrectly used instead of `user` for the username in the export filename, preventing potential runtime errors.
* Updated the assignment of `userRecord` to use the `user` object directly, ensuring correct insertion of user data into the export database.# Overview

Fixes: https://github.com/Termix-SSH/Support/issues/1020

### DIFF
--- a/src/backend/database/database.ts
+++ b/src/backend/database/database.ts
@@ -711,7 +711,7 @@ app.post("/database/export", authenticateJWT, async (req, res) => {
     }
 
     const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
-    const filename = `termix-export-${user[0].username}-${timestamp}.sqlite`;
+    const filename = `termix-export-${user.username}-${timestamp}.sqlite`;
     const tempPath = path.join(tempDir, filename);
 
     apiLogger.info("Creating export database", {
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
I checked this against the current `dev-2.6.1` and I owe you a straight answer: both problems you found are real, and both have since been fixed upstream — so merging this now would not add anything, and for one of the two files it would actually be a step backwards.

**`database.ts` — the `user[0]` / `user` mismatch.** Identical fix landed in #1111. Your diagnosis was correct.

**`PanePreview.tsx` — Tailscale host config.** This is the one worth explaining. The current base already builds `terminalHostConfig` with an explicit `authType` and a defensively derived port, but it resolves `host.sshPort ?? host.port` in **both** branches. Your version resolves the port only in the non-Tailscale branch:

```ts
// this PR
if (host.authType === "tailscale") {
  terminalHostConfig = { ...host, authType: "tailscale", instanceId };  // no port
}
```

Since Tailscale SSH can sit on a non-standard port — which your own comment calls out — taking this branch as-is would drop that resolution for exactly the case you were fixing. So the merge conflict here is not a formality; resolving it in this PR's favour would reintroduce the gap.

I did not push anything to your `main`, since this PR is from your fork's default branch and I did not want to touch it.

For what it is worth, you found two genuine bugs and reported them with a clear reproduction — that is the part that mattered, and the fixes exist now partly because the reports existed. If you want to keep contributing, `Termix-SSH/Support` has open issues that are not yet claimed, and working from a topic branch rather than `main` will save you this particular kind of collision next time.

Closing suggestion is yours to make — I would rather you decide than have a maintainer close it over you.

xyzulu:
Close away, I'm not precious with my thoughts etc. 

Thanks for all your contributions.

PR #1083, opened 2026-07-26
fix: SSH-login alerts silently dropped (channel load + auth middleware ordering)

brennanneoh:
Fixes [Termix-SSH/Support#1028](https://github.com/Termix-SSH/Support/issues/1028)

## Summary

Two bugs in the alert notification path, both making SSH-login alerts fail silently with nothing useful in the logs.

## Root cause

**Notification channel dropdown empty on mount.** Channels were only fetched when the user switched to the Channels tab, in that tab's own effect. Opening Edit Alert Rule without ever visiting the Channels tab first showed an empty channel picker even though channels existed, so a rule could be saved with none attached, with no indication anything was missing.

**`/internal/login-alert` registered after the global auth middleware.** This route is how the main backend tells the separate metrics-service process that an SSH login happened, so it authenticates via a loopback-IP check plus an `x-internal-auth` shared-secret token instead of a user JWT. It was registered after `app.use(authManager.createAuthMiddleware())`, so the JWT middleware rejected every call with `401 Missing authentication token` before the route's own check ever ran. `triggerLoginAlert` also never checked `res.ok` on its fetch, so the 401 vanished instead of surfacing anywhere.

With both bugs in play: alert rule configured correctly, channel configured correctly, but a real SSH login produced no notification and no error.

## Changes

- Load notification channels on `AlertsPanel` mount, not just after opening the channel dialog.
- Move the `/internal/login-alert` route registration to before the auth middleware, since it already self-authenticates by IP + shared secret. It's the only route moved; it still sits after CORS, cookie-parsing, JSON body-parsing, and the cache-control header middleware, just before `authManager.createAuthMiddleware()`.
- `triggerLoginAlert` now logs a warning when the metrics service responds with a non-OK status, instead of silently discarding the failure.

## User impact

Alert rules linked to a notification channel now actually fire notifications on SSH login. Future auth or delivery failures on this path will show up in the logs instead of disappearing.

## Testing

- Manually reproduced the original failure (empty channel dropdown, 401 from metrics service via a temporary debug log), applied the fix, and confirmed a real SSH login now delivers a notification end to end.
- `tsc --noEmit` clean.
- Added automated regression coverage:
  - `src/backend/tests/hosts/metrics/login-alert-route-order.test.ts` — asserts `/internal/login-alert` is registered before the auth middleware in source, since spinning up the full metrics-service app just for this route ordering felt out of scope.
  - `src/backend/tests/utils/alert-trigger.test.ts` — covers `triggerLoginAlert` warning on a rejected/non-OK response, staying silent on success, sending the expected payload, and warning (not throwing) when `fetch` itself rejects.
  - `src/ui/tests/sidebar/AlertsPanel.test.tsx` — covers channels loading on mount before the Channels tab is ever opened, the rule dialog's channel picker being populated without visiting that tab, and channels not being refetched on tab switch.

## Notes

Scoped to these two bugs. While debugging this I also found an unrelated bug in the same alert path's outbound fetch (`createDnsLookupHook` in `safe-outbound-fetch.ts` doesn't handle `lookupOptions.all` for Node's happy-eyeballs `autoSelectFamily`). Sending that as its own PR instead of bundling it here.


brennanneoh:
Added test coverage for both fixes: `AlertsPanel` channel-load-on-mount, `triggerLoginAlert`'s non-OK handling, and a source-order guard against the route being re-registered after the auth middleware.

That last one is worth flagging on its own. It checks the *position* of `app.post("/internal/login-alert", ...)` relative to `app.use(authManager.createAuthMiddleware())` in the source text, not actual runtime behavior. It'll catch someone moving the line back below the middleware, but it won't survive a refactor (route extracted into a router, code reformatted), and it says nothing about whether the middleware stack actually behaves correctly.

I looked at two ways to close that gap properly:

- **Real integration test.** Mock out the heavy dependencies in `metrics/index.ts` (DB layer, SSH clients, polling/connection managers), import the real file so the actual Express app and middleware stack get built, and use `supertest` to hit the route without a JWT and confirm it's rejected by the IP check (403) rather than the JWT middleware (401).
- **Extract the pattern.** Pull the loopback-IP + shared-secret check into its own small middleware function, and unit-test the ordering/composition directly, without touching Express at all.

Neither matches how this codebase currently handles this bug class though. `host-internal-routes.ts` has the identical shape (internal route with the same IP+token check, registered ahead of the JWT middleware in `host.ts`), and it has no test at all, not a source-order check, not an integration test. `supertest` also isn't a dependency anywhere in the repo; no backend test hits a real Express app.

So both options would be introducing a new testing convention rather than following an existing one, for a bug class the codebase currently just handles by getting the order right and leaving it untested. Given that, I left the lightweight source-order check as-is rather than pushing further in either direction. Happy to build out either A or B as a follow-up if you'd rather have real coverage here, just didn't want to invent a new pattern unasked-for in this PR.

ZacharyZcR:
Merged `dev-2.6.1` in to bring this up to date.

Some overlap to flag honestly: the route-ordering fix itself landed in #1100 while this was open, so `metrics/index.ts` now shows only your comment. But the rest of this PR is untouched by that and is the part I would not want to lose:

- **`AlertsPanel.tsx` — `loadChannels()` on mount.** #1100 did not touch the frontend at all. Without this the channel list is empty until something else happens to load it, which is the other half of what you titled "channel load". Genuinely separate bug, genuinely still broken before this.
- **`login-alert-route-order.test.ts`.** #1100 moved the route above the auth middleware but added nothing to stop it drifting back. Asserting the ordering directly is the right guard, and it is the kind of regression that is invisible until alerts silently stop again.
- **the comment on the internal endpoint** — it explains *why* the route sits before the middleware (IP + shared secret, not a user JWT). #1100 moved the code without recording the reason.
- the extra `alert-trigger` and `AlertsPanel` test coverage

Net against the current base: 5 files, +207 lines, all of it additive.

Verified: full suite 176 files / 1192 passed, lint 0 errors, build clean.

Two people fixing the same report from different ends — apologies that the backend half landed first.

PR #1084, opened 2026-07-26
fix: honor lookupOptions.all in custom DNS lookup hook

brennanneoh:
Fixes [Termix-SSH/Support#1027](https://github.com/Termix-SSH/Support/issues/1027)

## Summary

`safeOutboundFetch`'s custom DNS lookup hook used to always answer with a single `(address, family)` pair, even when the caller asked for the full address array via `lookupOptions.all`. Node's happy-eyeballs `autoSelectFamily` (the default since Node 20) is exactly that kind of caller, so any outbound request through this hook on Node ≥ 20 broke. I hit this through ntfy/webhook notification delivery, but it affects anything that goes through `safeOutboundFetch`.

**Update after rebasing onto `dev-2.6.1`:** the production fix in this PR turned out to be redundant — `dev-2.6.1` already carries an equivalent fix for the same bug (a `lookupOptions.all`-aware `fail` helper covering every exit path), landed independently. Behavior is the same, just written differently. This PR is now **test-only**: it adds the regression coverage described below, which wasn't present alongside the existing fix.

## Root cause

`net.connect` under `autoSelectFamily` calls a custom `lookup` hook with `{ all: true }` and expects the resolved address array back in the callback, not a single address. Answering the wrong shape doesn't raise a clean error, it corrupts undici's internal connect state, which only shows up two layers downstream as `fetch()` rejecting with a generic `TypeError: fetch failed`. The real reason, `TypeError [ERR_INVALID_IP_ADDRESS]: Invalid IP address: undefined`, is buried in `err.cause` and reads like a resolver bug rather than a callback-shape mismatch.

Confirmed with a standalone repro (stub `lookup` hook, real undici `Agent` + `fetch`, no network or DNS involved): calling `fetch()` against a plain loopback host still reproduces the exact same `fetch failed` / `ERR_INVALID_IP_ADDRESS` pair, which rules out DNS provider or network config as a factor.

## Changes

- No production changes — `dev-2.6.1`'s existing fix in `createDnsLookupHook` already answers every exit path (DNS error, empty result, blocked address, success) with the shape the caller asked for.
- Rewrote the `createDnsLookupHook` half of `src/backend/tests/utils/safe-outbound-fetch.test.ts` as flat `it.each` tables (matching the style already used in `termix-id-keys.test.ts`/`oidc-desktop-callback.test.ts`), parameterized over the three `lookupOptions` shapes a real caller can pass: `all: true`, `all: false`, and `all` omitted entirely.

## Test coverage added

- Success returns the full address list under `all: true`, the first address under `all: false`/omitted
- Rejects if any address is private, including an IPv4-mapped IPv6 spoof placed after a public address (not just checking `addresses[0]`)
- Distinct error when DNS returns no addresses
- Real DNS errors propagate untouched, in the shape the caller expects
- Underlying resolver is always called with `all: true` internally regardless of what the caller passed, so the blocklist always sees every candidate address

`tsc --noEmit` clean. Also previously confirmed end to end against a real `fetch()`/`Agent` call (not just the fake resolver above) using `docker/Dockerfile`'s exact base image, `node:26-slim`: before a fix, `safeOutboundFetch` fails with the `fetch failed`/`ERR_INVALID_IP_ADDRESS` pair described above; after, it succeeds.

## Notes

Scoped to this bug only, no blocklist changes. [Support#1027](https://github.com/Termix-SSH/Support/issues/1027) is the same one [#1079](https://github.com/Termix-SSH/Termix/pull/1079) flagged as a separate follow-up while fixing the SSRF blocklist false-positive in this same file.


ZacharyZcR:
Merged `dev-2.6.1` into this branch so it is up to date — the branch was behind and the base has moved a fair bit.

Heads-up on some overlap: the implementation half of this was fixed in #1101 (`safe-outbound-fetch.ts`), which landed while this was open. Your PR now shows as test-only against the new base — but that is not a loss, because the test work here is the part that stands on its own:

- `vi.fn()` on `fakeLookup` so the test can assert what the hook passes *down* to the resolver, not just what it hands back
- the `it.each` table covering all three `lookupOptions` shapes and their expected tail args
- the comment explaining that `lookupOptions.all` is the *caller's* expected callback shape and is independent of the internal `all: true` — that distinction is exactly what made the original bug easy to get wrong, and it was not written down anywhere

Verified on the new base: `safe-outbound-fetch.test.ts` 20 passed, full suite 174 files / 1192 passed, lint 0 errors, build clean.

Sorry the fix landed underneath you — the two were in flight at the same time. The tests are worth keeping on their own merit.

PR #1092, opened 2026-07-27
fix: preserve remote sync references

ZacharyZcR:
## Summary
- sync credentials and Vault profiles before dependent hosts and folders
- serialize cross-database relationships with stable sync IDs instead of local numeric IDs
- resolve stable references against the receiving database and reject missing dependencies

## Validation
- `npm test -- --run src/backend/tests/database/sync-references.test.ts`
- `npm run type-check`
- `npm run lint` (0 errors; existing warnings only)
- `npm run build`
- `git diff --check`

Closes Termix-SSH/Support#1034

### DIFF
--- a/electron/remote-sync.cjs
+++ b/electron/remote-sync.cjs
@@ -16,12 +16,12 @@ const fs = require("fs");
 const path = require("path");
 
 const SYNCED_ENTITY_TYPES = [
-  "hosts",
   "sshCredentials",
-  "sshFolders",
-  "snippets",
-  "snippetFolders",
   "vaultProfiles",
+  "sshFolders",
+  "snippetFolders",
+  "hosts",
+  "snippets",
   "dashboardServiceLinks",
   "homepageItems",
 ];
--- a/src/backend/database/routes/sync-references.ts
+++ b/src/backend/database/routes/sync-references.ts
@@ -1,0 +1,95 @@
+import type { SyncEntityType } from "../repositories/sync-tombstone-repository.js";
+
+export type SyncReferenceEntity = "sshCredentials" | "vaultProfiles";
+
+interface SyncReference {
+  field: string;
+  syncField: string;
+  entityType: SyncReferenceEntity;
+}
+
+const HOST_REFERENCES: SyncReference[] = [
+  {
+    field: "credentialId",
+    syncField: "credentialSyncId",
+    entityType: "sshCredentials",
+  },
+  {
+    field: "rdpCredentialId",
+    syncField: "rdpCredentialSyncId",
+    entityType: "sshCredentials",
+  },
+  {
+    field: "vncCredentialId",
+    syncField: "vncCredentialSyncId",
+    entityType: "sshCredentials",
+  },
+  {
+    field: "telnetCredentialId",
+    syncField: "telnetCredentialSyncId",
+    entityType: "sshCredentials",
+  },
+  {
+    field: "vaultProfileId",
+    syncField: "vaultProfileSyncId",
+    entityType: "vaultProfiles",
+  },
+];
+
+const REFERENCES: Partial<Record<SyncEntityType, SyncReference[]>> = {
+  hosts: HOST_REFERENCES,
+  sshFolders: [HOST_REFERENCES[0]],
+};
+
+export async function serializeSyncReferences(
+  entityType: SyncEntityType,
+  row: Record<string, unknown>,
+  resolveSyncId: (
+    entityType: SyncReferenceEntity,
+    id: number,
+  ) => Promise<string | null>,
+): Promise<Record<string, unknown>> {
+  const serialized = { ...row };
+  for (const reference of REFERENCES[entityType] ?? []) {
+    const id = serialized[reference.field];
+    serialized[reference.syncField] =
+      typeof id === "number"
+        ? await resolveSyncId(reference.entityType, id)
+        : null;
+    delete serialized[reference.field];
+  }
+  return serialized;
+}
+
+export async function deserializeSyncReferences(
+  entityType: SyncEntityType,
+  row: Record<string, unknown>,
+  resolveId: (
+    entityType: SyncReferenceEntity,
+    syncId: string,
+  ) => Promise<number | null>,
+): Promise<Record<string, unknown>> {
+  const deserialized = { ...row };
+  for (const reference of REFERENCES[entityType] ?? []) {
+    const syncId = deserialized[reference.syncField];
+    delete deserialized[reference.syncField];
+    delete deserialized[reference.field];
+
+    if (syncId == null) {
+      deserialized[reference.field] = null;
+      continue;
+    }
+    if (typeof syncId !== "string") {
+      throw new Error(`Invalid ${reference.syncField}`);
+    }
+
+    const id = await resolveId(reference.entityType, syncId);
+    if (id === null) {
+      throw new Error(
+        `Missing ${reference.entityType} dependency ${reference.syncField}=${syncId}`,
+      );
+    }
+    deserialized[reference.field] = id;
+  }
+  return deserialized;
+}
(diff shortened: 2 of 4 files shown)


kacperpietrzyk:
I added the macOS analysis on Support#1034, so I read this branch against `main` @ `4a7117b` with that in mind. The approach here is the right one — serialize cross-database relationships by `syncId`, drop the local numeric ids, order dependencies before dependents. Two findings below: the first clears a concern I had, the second is a gap I think is worth closing before merge.

## Reference coverage is complete

I cross-checked `REFERENCES` in `sync-references.ts` against every table in `SYNCED_ENTITY_TYPES`:

| table | numeric FKs (excl. `userId`) | covered |
|---|---|---|
| `hosts` | `credentialId`, `rdpCredentialId`, `vncCredentialId`, `telnetCredentialId`, `vaultProfileId` | yes |
| `sshFolders` | `credentialId` | yes |
| `sshCredentials`, `snippets`, `snippetFolders`, `vaultProfiles`, `dashboardServiceLinks` | none | n/a |
| `homepageItems` | `folderId` | see below |

The `folder` columns on hosts/snippets/credentials/vaultProfiles are `text` folder *names*, so they are portable across databases and correctly left alone.

`homepageItems.folderId` is a bare `integer` with no FK constraint — exactly the shape that would corrupt silently rather than raise. But it appears to be dead: declared in `schema.ts:1153` and in the raw DDL at `db/index.ts:2298`, and read or written by no application code (a whole-repo grep for `folderId|folder_id` returns nothing in the homepage layer). Not a problem today; only worth revisiting if that column is ever wired up.

## The failure amplifier from #1034 is still present — and this PR adds new ways to reach it

Support#1034 had two compounding defects: the numeric FK copied verbatim (fixed here), and the sync loop turning any single failure into a total, permanent freeze. The second is untouched by this branch:

- In `electron/remote-sync.cjs`, the `for (const entityType of SYNCED_ENTITY_TYPES)` loop (L255) has no per-entity `try`/`catch`; the only `try` (L251) wraps the entire loop.
- `writeJson(getRemoteSyncStatePath(), state)` (L288) is reached only after the loop runs to completion. Any throw jumps to the catch at L303, so state is never persisted and `since` never advances — including for the entity types that *did* sync successfully earlier in the same pass. The `sawAuthFailure` early return (L279–L286) discards state the same way.

That matters more after this change rather than less, because `deserializeSyncReferences` introduces two new throw sites:

```js
throw new Error(`Invalid ${reference.syncField}`);
throw new Error(
  `Missing ${reference.entityType} dependency ${reference.syncField}=${syncId}`,
);
```

These run server-side, so to be precise about the path — a host referencing a credential that was deleted (tombstoned) on the other side fails like this:

`deserializeSyncReferences` throws (`sync.ts:311`) → caught at `sync.ts:355` → **HTTP 500 `Failed to upsert row`** on `POST /sync/<entity>` → the client throws on the failed request → `syncEntity`'s own catch rethrows it (`remote-sync.cjs:486` — it converts only `authFailure` into a normal return) → the outer catch at L303, with `state` unpersisted.

That is exactly the signature reported in #1034: `Sync error: Request failed (500): http://127.0.0.1:30001/sync/hosts`. Reordering dependencies first makes this rarer, not impossible.

Suggested remedy, in keeping with the size of this PR: wrap the loop body in `try`/`catch`, record the per-entity error, continue to the next entity type, and persist `state` for the types that succeeded. Per-entity and inside the loop — not a job-level skip.

## Not verified yet

I have not confirmed at runtime that hosts bind to the *correct* credential after a real sync, checked by `syncId` mapping rather than by the absence of a 500. That is the check worth insisting on here, because the naive fix for #1034 (preloading the server's credentials locally) goes green while binding every host to the wrong credential — `ssh_credentials.id` is `AUTOINCREMENT`, so a freed id is never reissued and the rows land shifted.

I have this repo building locally and a 2.6.0 server + desktop pair to test against. Happy to run this branch end-to-end and report the `syncId` mapping back here if that would help.

kacperpietrzyk:
Ran this branch end to end, as offered above. Summary: **the fix is correct**, one of my two review
points was wrong and I'm retracting it, and the run surfaced a separate latent defect that this PR
makes reachable.

Setup: server built from `dev-2.6.1` @ `e3205ec` (`docker/Dockerfile`), isolated volume, port 8081.
Client is this branch's `electron/remote-sync.cjs`, driven headless through `app.whenReady()` with
`TERMIX_DATA_DIR` pointed at a throwaway profile, one deliberate `syncNow()` per run — no UI, no
shared state with an installed app.

The fixture reproduces the Support#1034 id geometry rather than a clean profile, because a clean
profile cannot fail: with no local rows, every server credential lands at a fresh id and resolves
correctly under the naive preload workaround too. So the local side holds one local-only credential
occupying `id=1`, and the server's hosts reference credentials `id=1..3`.

## Verified by `syncId` mapping, not by the absence of a 500

| host | `credentialSyncId` on the wire | server credential | local credential after sync |
|---|---|---|---|
| box-alpha | `f2186284…` | id 1 `srv-alpha` | **id 2** `srv-alpha` |
| box-bravo | `82629ab4…` | id 2 `srv-bravo` | **id 3** `srv-bravo` |
| box-charlie | `2d0c4ff9…` | id 3 `srv-charlie` | **id 4** `srv-charlie` |

Local ids are shifted by one against the server, and every host still binds to the right
credential — which is the discriminating result: id-based copying would have bound `box-alpha` to
the local `id=1` impostor. `credentialId` is gone from the wire entirely, the local-only credential
pushed up as a new row, and `sync-references.test.ts` + `backend-paths.test.ts` pass (7/7).

## Retracting my second review point

I claimed a host referencing a tombstoned credential would hit `Missing <entity> dependency` → 500
→ frozen cycle. **That trigger doesn't exist**: `DELETE /credentials/:id` clears the referencing
hosts server-side, so the host is serialized with `credentialSyncId: null` and never carries a
dangling reference. My description of the loop is still accurate as code — one `try` around the
whole `for`, `writeJson(state)` only after a complete pass — but I couldn't reach the throw, and
the example I gave to justify it was wrong. Sorry for the noise.

## What the run did surface

After a completed pass, sync stops transferring anything at all — filed as
Termix-SSH/Support#1050. Short version: `GET /sync/:entityType?since=…` compares
`gt(table.updatedAt, since)` on a TEXT column, where the column holds `CURRENT_TIMESTAMP` output
(`2026-07-29 10:11:21`) and `since` is `new Date().toISOString()`
(`2026-07-29T10:06:55.172Z`). Lexically `' '` < `'T'` at position 10, so the comparison is false
regardless of the instant; same for the tombstones endpoint via `listSince`.

```
GET /sync/hosts                                    -> 3 rows (newest 2026-07-29 10:11:21)
GET /sync/hosts?since=2026-07-29T09:00:00.000Z     -> 0 rows      <- format the client sends
GET /sync/hosts?since=2026-07-29 09:00:00          -> 3 rows      <- format the column stores
```

Not caused by this PR — that code is unchanged since #1085. It was *masked* by Support#1034: the
loop threw before persisting state, so `since` never left `null` and every cycle was a full sync.
Now that a pass completes, passes 2..n pull nothing and report success, so edits and deletions stop
propagating both ways with no error anywhere. Worth fixing before 2.6.1 ships, since that's when
users will meet it. I have the repro standing and can send a patch if it's wanted.

kacperpietrzyk:
Closing the loop on the half of my review I kept: the failure amplifier is real, and I've now hit it
for real rather than hypothetically.

The trigger isn't the dangling reference I guessed at — that one I retracted above and it stays
retracted. It's `POST /sync/tombstones`, which is registered after `POST /:entityType` and so gets
matched by the wildcard: `"tombstones"` passes as an `:entityType`, fails `isValidEntityType`, and
returns 400. My harness run:

```
lastError: "Request failed (400): http://127.0.0.1:30001/sync/tombstones"
```

That 400 propagates out of `syncEntity`, past the single `try` at the top of the loop, and `state`
is never written — so the entity types that had already synced successfully earlier in the same
pass lose their cursors too, exactly as described. So the structural point stands on its own
evidence now, with a trigger that reproduces, and the per-entity `try`/`catch` remains worth having
independently of any single cause.

Both underlying defects are filed as Termix-SSH/Support#1051 with a patch in #1139; the cursor
issue that hid them is Support#1050 / #1138.

PR #1108, opened 2026-07-27
feat: add host export dialog with host and field selection

maxiwolleb:
# Overview

Replaces the two export items in the Hosts panel (`Export All` / `Export for Sharing`) with a single `Export...` dialog where you pick the hosts, pick the fields, toggle credentials, and see the actual file before it downloads.

- [x] Added: `Export...` dialog with All/Selected scope, field-group selection, credential toggle and a live preview
- [x] Added: `Export` button in the selection-mode bulk bar, which opens the dialog pre-filled with what you ticked
- [x] Removed: `Export All` and `Export for Sharing` menu items (both dropdowns, desktop and mobile)
- [x] Fixed: two nested secrets that `?share=1` does not scrub (details below)

Closes the request in Termix-SSH/Support#1001.

# Changes Made

The credential decision stays on the server. Unticking Credentials makes the client call the existing `GET /host/db/hosts/export?share=1`; ticking it calls the normal one. Host and field filtering happen client-side on whatever came back. **No backend changes.**

Preview and the downloaded file come from the same `buildExportPayload()` call — the preview is literally the file, with secret *values* masked as `"<included>"` so you can look at it (or screen-share it) without exposing anything. They can't drift apart.

New files:
- `src/ui/sidebar/host-export-payload.ts` — pure shaping logic, no React
- `src/ui/sidebar/HostExportDialog.tsx` — the dialog
- two test files, 23 tests

`HostsPanel.tsx` gets smaller: `handleExportHosts` is gone.

Only `en.json` was touched for strings — nothing under `locales/translated/`.

## Heads-up on nested secrets

While building the preview I found that `?share=1` nulls the top-level secrets but not every nested one. Two come back in plaintext:

- `guacamoleConfig["gateway-password"]`
- `socks5ProxyChain[].password`

I confirmed this against a running instance, not just by reading the code. For RDP/VNC hosts `guacamoleConfig` is the only secret-bearing field, so a "share" export could carry a gateway password while the UI said credentials were excluded.

This PR nulls both client-side before they can reach the file, and there are tests that fail if that nulling is removed. But the client is the wrong place for it long-term — the export route in `src/backend/database/routes/host.ts` builds its redaction as a hand-maintained deny-list (`shareMode ? null : x` per field), so any new field is exported by default until someone remembers. Flipping that to an allow-list would fix this properly and stop it recurring. Happy to do that as a follow-up PR if you want it, I just kept this one to the UI so it stays reviewable.

## Two things worth flagging

**Default field groups are all on.** Since `Export All` is gone, I made every field group tick by default (credentials stays off) so the default export isn't lossier than the button it replaces.

**Host identity uses a tuple.** The export payload has no host `id`, so the dialog matches your ticked hosts on `(name, ip, port, username, connectionType)`. Two hosts identical across all five get exported together. There's a test pinning that. Adding `id` to the export payload would fix it properly — one line in the backend — but that felt like your call, so I left it out.

# Related Issues

- Related to Termix-SSH/Support#1001

# Screenshots / Demos

Desktop and mobile below. Tested against a real instance with 5 hosts and a shared credential: All/Selected scope, orphaned credentials getting pruned when you deselect the host using them, `share=1` present/absent on the right toggles, and the downloaded file matching the preview.
image
image

# Checklist

---

Squash-merge is fine by me, it's one commit anyway.

Thanks for building Termix, and for the export feature in the first place — I wanted to take a shot at improving it myself rather than just opening another request. If you want anything changed, split up, or done differently, say the word and I'll do it. I just like the project and wanted to help.


ZacharyZcR:
I retargeted this from `main` to `dev-2.6.1`. `main` only moves on a release — all development lands on the current dev branch, which is 58 commits ahead of `main` right now. A PR based on `main` also cannot close its linked issue automatically, since that only fires on the default branch.

Good news: it still merges cleanly against the dev branch, so nothing is needed from you for the retarget itself.

One thing to know when you look at the checks: `package.json` on `dev-2.6.1` was invalid JSON from 2026-07-29 until today — the version bump dropped a closing quote, so every CI run on that base failed before reaching a single test, with an opaque rolldown error that named the wrong line. Fixed in #1147. If this PR shows a red or stale run from the last two days, that is why; a fresh run should be clean.

Not a review — I have not read the change yet.

LukeGus:
LGTM after review, thanks.

PR #1116, opened 2026-07-27
fix: use jump host SOCKS proxy settings

ZacharyZcR:
## Summary
- source SOCKS5 settings exclusively from the first jump host
- stop applying destination host proxy settings to the first jump connection
- cover direct, proxy, and serialized proxy-chain behavior with regression tests

## Validation
- `npx vitest run src/backend/hosts/jump-host-proxy.test.ts`
- `npm run type-check`
- ESLint on changed files
- `npm run build`

Closes Termix-SSH/Support#894


ZacharyZcR:
Rebased onto `dev-2.6.1` after #1115 landed. The two overlapped in `guacamole/routes.ts` and the resolution is worth spelling out, because taking either side wholesale would have been wrong:

- **#1115** added the guacd-host resolution and `resolveJumpTunnelEndpoint` for containerised guacd
- **#1116** removes the `proxyConfig` argument from `createJumpHostChain`, so the first hop is dialled through *its own* SOCKS5 settings rather than the target host's — which is the actual point of Support#894

Kept both: the tunnel-endpoint resolution stays, the inlined `proxyConfig` construction and its `ProxyNode`/`SOCKS5Config` imports are gone, and the call is now `createJumpHostChain(jumpHosts, userId)`.

Verified after the merge:

- `npm test` — 169 files, 1165 passed, 1 skipped
- `npm run type-check`, `npm run build`, `npm run lint` (0 errors)
- `jump-host-proxy.test.ts` — 3 passed

PR #1117, opened 2026-07-27
Fix OIDC login when the ID token is not a verifiable JWS

ZacharyZcR:
## Summary

- reject a non-JWS `id_token` with a distinct `OIDCTokenFormatError` before any JWKS lookup, so the shape check is separable from a real verification failure
- fall back to the userinfo endpoint when the ID token carries no verifiable claims, instead of failing the login with `JWSInvalid: Invalid Compact JWS`
- keep rejecting the login when signature or claim verification actually fails, and keep rejecting unverifiable back-channel logout tokens
- report an encrypted (JWE) ID token with an actionable message naming the provider setting to change
- add regression coverage for JWE and other malformed token shapes

`jwtVerify` throws `JWSInvalid` when a token does not split into exactly three
segments — a shape check, not a signature check. Authentik issues an encrypted
five-segment JWE `id_token` when the provider has an encryption key configured,
so verification never succeeded for those deployments. 2.5.0 hid this behind a
catch-all that decoded the unverified payload; for a JWE its `parts.length === 3`
guard was false, `userInfo` stayed null, and the login completed via userinfo.
Removing that fallback stopped Termix trusting unsigned claims, but also turned
the pre-existing failure into a hard login failure. This restores the 2.5.0
outcome without the unsigned-claim decode.

## Validation

- `npm test -- --run src/backend/tests/database/routes/user-oidc-utils.test.ts` (27 passed)
- `npm run type-check`
- `npm run lint` (0 errors; 44 existing warnings only)
- `npm run build`
- `git diff --check`

Closes Termix-SSH/Support#1016
Closes Termix-SSH/Support#1018

### DIFF
--- a/src/backend/database/routes/user-oidc-utils.ts
+++ b/src/backend/database/routes/user-oidc-utils.ts
@@ -10,6 +10,17 @@ import {
 const BACKCHANNEL_LOGOUT_EVENT =
   "http://schemas.openid.net/event/backchannel-logout";
 
+/**
+ * Raised when a token cannot be verified because it is not a compact JWS,
+ * as opposed to a signature or claim check that actually failed.
+ */
+export class OIDCTokenFormatError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "OIDCTokenFormatError";
+  }
+}
+
 function normalizeIssuer(url: string): string {
   return url.trim().replace(/\/+$/, "");
 }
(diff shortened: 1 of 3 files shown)


ZacharyZcR:
Rebased to be independent of #1102 — the two touched the same lines, and both overlaps were avoidable on this side.

- `user-oidc-utils.ts`: dropped a cosmetic change (reusing the already-split `segments[0]` for the header decode). #1102 replaces that line with `decodeProtectedHeader`, so this PR now leaves it alone entirely. The segment-count guard sits above it and is untouched by either.
- the test file: this suite now imports the module as a namespace instead of adding to the shared destructuring block at the top, which #1102 also extends.

Verified both merge orders against `dev-2.6.1`:

```
#1102 → #1117   auto-merged, 28 tests pass, type-check clean
#1117 → #1102   auto-merged, 28 tests pass
```

So these two can be merged in either order with no manual resolution.

PR #1134, opened 2026-07-28
Groundwork for Postgres and MySQL backends

ZacharyZcR:
**Draft — foundation only.** Nothing is wired up, SQLite remains the only runtime path, and no existing behaviour changes. Opening it early so the approach can be argued about before 52 tables are touched.

## What this is

#1127 made the repository layer dialect-agnostic (46/46 repositories are drizzle-only). This adds the pieces needed to actually target a second engine:

- **`DatabaseDialect`** — `sqlite | postgres | mysql`, resolved from `DATABASE_DIALECT`, defaulting to `sqlite`. An unknown value throws rather than silently falling back.
- **`column-kit.ts`** — the per-dialect type decisions in one file
- **`schema-portable.ts`** — `settings` and `users` declared for all three dialects as a proof slice
- **`pg` and `mysql2`** as dependencies

## The differences the kit absorbs

Verified by test, not assumed:

| | sqlite | postgres | mysql |
|---|---|---|---|
| `bool("is_admin")` | `integer` | `boolean` | `boolean` |
| `shortText("id")` (a key) | `text` | `varchar(255)` | `varchar(255)` |
| identifier quoting | `"users"` | `"users"` | `` `users` `` |
| placeholders | `?` | `$1` | `?` |
| boolean param value | `1` | `true` | `true` |

MySQL cannot index unbounded `TEXT`, which is why key columns need `shortText` rather than `text` — that one is easy to get wrong 52 times.

## What I tried first, and why it is not that

The obvious design is one declaration per table, materialised per dialect:

```ts
function defineUsers(kit: ColumnKit) {
  return kit.table("users", { id: kit.shortText("id").primaryKey(), ... });
}
```

**It does not type-check.** `sqliteTable`, `pgTable` and `mysqlTable` have incompatible signatures, so `kit.table` is `TS2349: This expression is not callable`, and forcing it collapses the column types to `any` — losing exactly the inference the repositories rely on.

So the bodies are declared per dialect. The kit still earns its place by keeping the *type choices* in one file, leaving the per-dialect bodies mechanical.

**This is the open question for the rest of the schema.** Three hand-maintained copies of 52 tables is not something I would sign up for. Options, roughly in order of my preference:

1. **Generate** the three schema modules from one declaration at build time — keeps a single source of truth and full inference, adds a codegen step
2. **Postgres only**, skip MySQL — halves the problem; worth asking whether MySQL is actually wanted
3. Hand-maintain three copies with a test asserting they stay structurally identical — cheap to start, expensive forever

I would like a steer before going further.

## Deliberately not here

- **No connection wiring.** `db/index.ts` still opens SQLite unconditionally.
- **No DDL.** That file has 67 hand-written `CREATE TABLE` and 122 `addColumnIfNotExists` calls, all SQLite. Porting them is a separate piece of work, and probably the moment to adopt drizzle-kit migrations instead.
- **No storage-model change.** The database is loaded into memory and serialised back to an encrypted file; 17 files reference `DatabaseSaveTrigger` and 43 repositories take an `onWrite` hook. On Postgres that whole mechanism is meaningless and encryption has to be reconsidered (column-level, or delegated to the server). This is the real design work and it deserves its own discussion.

## Worth being explicit about

**This is multi-backend, not a migration.** The desktop app embeds its backend and cannot ship a database server, so SQLite has to keep working. Anything that assumes a single engine is wrong.

## Validation

- new `multi-dialect.test.ts` — 11 cases: dialect resolution incl. rejecting unknown values, structural equivalence across the three schemas, per-dialect SQL type mapping, and real query building for all three engines (no server needed — the builders are exercised via drizzle's driverless constructors)
- `npm test` — 177 files, 1209 passed, 1 skipped, 0 failed
- `npm run lint` (0 errors), `npm run build`, `npm run format:check`


ZacharyZcR:
## Update: constraints verified

The first slice only covered plain columns, which was not enough to trust the approach. The real schema also has **92 foreign keys** (80 `cascade`, 12 `set null`) and **14 unique columns** — if those did not port, the whole thing would be dead on arrival.

Added `audit_logs` and `ssh_folders` to the proof slice, chosen to cover the remaining constructs:

| construct | covered by |
|---|---|
| nullable FK, `ON DELETE SET NULL` | `audit_logs.user_id` — the one keeping the trail after a user is deleted (#1132) |
| required FK, `ON DELETE CASCADE` | `ssh_folders.user_id` |
| unique column | `ssh_folders.sync_id` |
| autoincrement surrogate key | both |

**All of it holds on all three dialects.** The surrogate key is the interesting one — it is spelled three different ways underneath:

```
sqlite   integer primary key autoincrement
postgres serial
mysql    int auto_increment
```

and the column kit hides that, so the schema bodies stay mechanical.

Also worth flagging for whoever picks this up: **`getTableConfig` is dialect-specific**. Calling the sqlite one on a pg table does not throw a useful error — it fails with `Cannot convert undefined or null to object`. The test now imports all three.

### Where this leaves the open question

The constructs are no longer in doubt; what remains is purely the maintenance question from the PR description. With foreign keys included, the per-dialect bodies are still mechanical but now cross-reference each other (`pgPortable.users.id` inside `pgPortable.auditLogs`), which makes hand-maintaining three copies of 52 tables meaningfully worse — a rename has to land in three places consistently or the FK silently points at the wrong table.

That pushes me further toward generating the three modules from one declaration. Still happy to go whichever way you prefer, but I would rather not hand-write 150 table definitions before that is decided.

Current state: 180 files, 1245 passed, lint 0 errors, build clean. Still draft, still nothing wired up.

ZacharyZcR:
## Update: all 52 tables, generated

The open question in the description was how to handle the rest of the schema. I went with generation, and it worked out better than expected.

**`schema.ts` stays the single source of truth.** `schema.pg.ts` and `schema.mysql.ts` are produced by `scripts/generate-dialect-schema.cjs`:

```
schema.ts          1193 lines   52 tables
schema.pg.ts       1201 lines   52 tables
schema.mysql.ts    1200 lines   52 tables
```

All three compile. **0 type errors** across the whole tree.

The hand-written proof slice and the column kit are gone — the generator now holds those decisions, so there is one place to look rather than three files plus a kit.

### What it transforms

Exactly the closed set the kit had enumerated:

| | postgres | mysql |
|---|---|---|
| `integer(…, { mode: "boolean" })` | `boolean` | `boolean` |
| `integer(…).primaryKey({ autoIncrement })` | `serial` | `int().autoincrement()` |
| `integer` | unchanged | `int` |
| `real` | `doublePrecision` | `double` |
| key-bearing `text` | `varchar(255)` | `varchar(255)` |

"Key-bearing" means primary key, unique, **or either end of a foreign key** — both sides have to agree or the constraint will not create. That is the rule most likely to be got wrong 92 times by hand.

One thing worth recording: the first run left 7 compile errors, all from `integer(…, { mode: "boolean" })` declarations that Prettier had wrapped across lines. A single-line regex missed them. That is precisely the class of mistake that would have been silent in a hand-written port — here the compiler caught it immediately, which is an argument for generation on its own.

### Drift protection

`npm run lint` now runs `--check` first, so editing the schema without regenerating fails CI:

```
$ npm run schema:check
[generate-dialect-schema] src/backend/database/db/schema.pg.ts is out of date
[generate-dialect-schema] run `node scripts/generate-dialect-schema.cjs` and commit the result
```

Verified by deliberately adding a column and confirming it fails, then reverting.

### Tests

- `generate-dialect-schema.test.ts` — 16 cases over a fixture covering every construct, including the multi-line boolean form, determinism, and that foreign-key behaviour passes through verbatim
- `multi-dialect.test.ts` — 13 cases now running against the **real generated schemas**: same table set across dialects, per-dialect column types, autoincrement spelled three ways, both `cascade` and `set null` preserved, nullability and uniqueness intact, and real query building for all three engines
- `npm test` — 181 files, 1259 passed, 1 skipped
- lint 0 errors, build clean

### Still not wired up

`db/index.ts` continues to open SQLite unconditionally. Next is the storage model — `DatabaseSaveTrigger` and the encrypted-file round trip have no meaning on a client-server engine, and 43 repositories take an `onWrite` hook that exists only to serve it. That is the last structural piece before a connection can actually be opened.

ZacharyZcR:
## Update: durability split, and the backends documented

Two more pieces, which together were the last structural work before a connection can be opened.

### Durability

The `onWrite` hook every repository takes exists for one reason: SQLite here runs in memory and has to be serialised back to its encrypted file. On Postgres and MySQL a committed write is already durable.

The fix turned out to be one function rather than 43. The factory decides:

```ts
export function createCurrentRepositoryWriteHook(
  reason: string,
): (() => Promise<void>) | undefined {
  if (!needsExplicitPersist(resolveDatabaseDialect())) return undefined;
  return () => DatabaseSaveTrigger.forceSave(reason);
}
```

No hook is installed at all, rather than one that does nothing — repositories already call it as `this.onWrite?.()`, so **none of the 43 repositories change**.

### docs/database-backends.md

Mostly to be explicit about encryption, since that is the part most likely to be misread — and where I would rather over-communicate than let someone assume parity that does not exist.

**Identical on all three engines:** field-level encryption. Every credential — host passwords, private keys, TOTP secrets, vault certs, per-recipient shared secrets — is encrypted in the application before it reaches the database. SSO secrets under the system key (#1135). This is the protection that matters most and it does not change.

**Different:** whole-file encryption has no equivalent on a client-server engine. So on Postgres/MySQL these become readable by anyone with database access, where SQLite covered them:

- host names, addresses, ports, usernames
- folder and snippet names, and snippet contents
- audit log entries
- session recording metadata
- user names, roles, API key hashes

None are credentials. Together they describe the estate. The doc says plainly that encryption at rest on those engines is the operator's job — TDE, encrypted volume, encrypted filesystem — and that **backups are plain unless you encrypt them**.

There is a side-by-side threat model table. The honest summary is that field-level encryption holds everywhere; the difference is entirely in the stolen-volume and backup rows.

### State

`npm test` — 181 files, 1260 passed, 1 skipped. Lint 0 errors, build clean, schema drift check green.

Still draft, still nothing wired up. What is left before `DATABASE_DIALECT=postgres` does anything:

1. **DDL** — 67 hand-written `CREATE TABLE` and 122 `ADD COLUMN` migrations, all SQLite. This is the moment to adopt drizzle-kit migrations rather than port them by hand.
2. **Connection opening** in `db/index.ts`.
3. **`getCurrentSettingValue`** — a synchronous settings read used during startup; it needs an async path first.

(1) is the big one and I would rather agree on the migration approach before starting it.

ZacharyZcR:
## Update: DDL generated, settings unblocked — connection still open

Two of the three remaining blockers are done. The third turned into a design question rather than a task, so I stopped rather than guess.

### DDL — drizzle-kit, not a hand port

`db/index.ts` hand-writes **67 `CREATE TABLE`** and **122 `ADD COLUMN`** migrations, all SQLite. Porting those by hand for two more engines was never going to be right.

drizzle-kit generates them from the schema modules instead:

```
drizzle/postgres/0000_*.sql   817 lines
drizzle/mysql/0000_*.sql      869 lines
```

The type mapping is already correct, because the schemas it reads are themselves generated:

```sql
CREATE TABLE "alert_firings" (
	"id" serial PRIMARY KEY NOT NULL,
	"user_id" varchar(255) NOT NULL,
	"acknowledged" boolean DEFAULT false NOT NULL
```

`npm run schema:migrations` regenerates both. This also gives the project a real migration story, which is worth having regardless of the multi-backend work — the current 122 `addColumnIfNotExists` calls are not one.

### Settings — a synchronous read that had to stay synchronous

27 call sites read settings synchronously: during startup, inside request handlers, and from the guacd bootstrap. better-sqlite3 allows that; Postgres and MySQL have no synchronous query at all.

Making all 27 async would have pushed `await` through code paths that have no reason to be asynchronous — including `readGuacdOptions`, which runs while constructing the guacd server. So instead:

- settings are a handful of low-cardinality rows, changed rarely and read constantly, so the whole table is cached
- primed once at startup; `SettingsRepository` updates the cache on `set`, `delete` and `deleteLike` (which has `returning`, so the deleted keys are known)
- **SQLite still reads the database directly** and stays authoritative — the cache only serves engines that cannot be read synchronously

Nine tests cover the semantics that matter: a read before priming returns null rather than throwing (startup ordering), a write is visible immediately, and writes while unprimed are ignored rather than half-populating a cache that readers could not distinguish from a real one.

### What stopped me on the connection

```ts
export interface DatabaseContext {
  drizzle: BetterSQLite3Database<typeof schema>;   // ← bound to one engine
}
```

**43 repositories depend on that inference.** The three drizzle instance types are not interchangeable — the same incompatibility that killed the single `define(kit)` helper earlier in this PR.

The options I can see:

1. **Widen to a minimal common interface** — keeps type safety where it is used, but has to be written and maintained by hand against drizzle's API surface
2. **Generic over the dialect** — `DatabaseContext<D>` threaded through 43 repositories; correct, invasive
3. **`any` / assertion at the boundary** — one line, and throws away the inference the repositories were built on

I lean towards (1), scoped to what the repositories actually call — which after #1127 is a closed set, since none of them can reach the raw driver any more. But that is a decision worth making deliberately rather than in passing, so I have left it.

### State

`npm test` — 182 files, 1269 passed, 1 skipped. Lint 0 errors, build clean, schema drift check green. Still draft; `DATABASE_DIALECT=postgres` still does nothing.

ZacharyZcR:
## Update: the RETURNING gap, and the connection layer

### MySQL's real blocker was not the connection

MySQL has no `RETURNING` clause, and drizzle's mysql-core does not expose the method. **156 call sites here read the result of a write.** That, not connection handling, is what stood between this and MySQL support.

Classifying them changed the picture:

| what the caller needs | count | on MySQL |
|---|---|---|
| just `.length` | **92** | `affectedRows` — **no change to the query** |
| first row, mostly an autoincrement id | 30 | `insertId` |
| the rows themselves | 17 | needs read-then-write |
| other | 17 | case by case |

**59% are satisfied by reading the result differently, not by writing different SQL.** `mutation-result.ts` provides `rowsAffected()` and `insertedId()`, which understand both the returned-array shape and the MySQL header.

One detail worth recording: **mysql2 returns `[ResultSetHeader, fields]`, which is itself an array**, so "is it an array" cannot distinguish it from a `returning()` result. The header has to be identified by its own fields. My first version got this wrong and the test caught it.

The remaining ~34 sites genuinely consume returned rows. Those need a read inside a transaction to stay correct under concurrency, so they will be handled individually rather than behind a helper that quietly adds a round trip. `supportsReturning()` is the seam.

### The type problem, and what I decided

drizzle's three `Database` classes share no base class:

```
BaseSQLiteDatabase<TResultKind, TRunResult, …>
PgDatabase<TQueryResult, …>
MySqlDatabase<TQueryResult, TPreparedQueryHKT, …>
```

There is no honest type covering all three. A union is not callable — the same incompatibility that killed the `define(kit)` helper earlier. A generic would have to be threaded through 43 repositories and every method on them.

So `DatabaseContext.drizzle` is now `PortableDatabase`, **still the SQLite type underneath**. I want to be straight about this: it is an approximation, and at runtime the handle may be a Postgres or MySQL instance.

What makes it defensible rather than sloppy:

- it is **named and documented** as such, instead of a bare `BetterSQLite3Database` that reads as a guarantee
- the equivalence is **asserted** in `multi-dialect.test.ts` — identifier quoting, placeholder style, boolean coercion — not assumed
- the one place the surfaces genuinely differ is RETURNING, and that is handled explicitly rather than hidden

If you would rather pay for full type safety, option (2) from my last comment — generic over the dialect — is still open. It is invasive but not impossible, and I would rather be told than assume.

### connect.ts

Opens Postgres and MySQL from `DATABASE_URL`. The driver and schema module are imported lazily, so a SQLite deployment loads neither.

The URL scheme is validated against the configured dialect first:

```
DATABASE_URL is a "mysql://" URL but DATABASE_DIALECT is "postgres".
Expected one of postgres://, postgresql://.
```

Without that check the mismatch surfaces as a driver error deep in a stack that never mentions the real problem. `mariadb://` is accepted for MySQL since it speaks the same protocol.

### State

184 files, 1298 passed, 1 skipped. Lint 0 errors, build clean, format clean, schema drift green.

`db/index.ts` still opens SQLite unconditionally — wiring `connect.ts` into it is next, and it is the first change in this PR that will actually alter runtime behaviour.

ZacharyZcR:
## Wired up: `DATABASE_DIALECT` now takes effect

This is **the first commit in this PR that changes runtime behaviour.** Everything before it was additive.

`initializeDatabase()` forks on the dialect:

```ts
const dialect = resolveDatabaseDialect();
if (dialect !== "sqlite") {
  await initializeRemoteDatabase(dialect);
  return;
}
await initializeCompleteDatabase();
await handlePostInitFileEncryption();
```

`resolveDatabaseDialect()` defaults to `"sqlite"` and throws on an unrecognised value, so **an existing deployment takes the identical path it always did** — the branch is only reachable by setting the variable.

### The remote path is short, and that is the point

```ts
db = await connectRemoteDatabase(dialect);
await runRemoteMigrations(dialect, db);
await primeCurrentSettingsCache();
```

Most of what the SQLite path does has no counterpart: no file to decrypt, no in-memory copy to keep in step with disk, no inline DDL. What does carry over is **priming the settings cache** — 27 call sites read settings synchronously, which better-sqlite3 allows and no remote driver does. Doing it here rather than in `starter.ts` means it cannot be forgotten.

That import is dynamic, deliberately: `factory.ts` imports `getDb` from `db/index.ts`, so a static import would close the cycle at module-load time.

### Schema comes from drizzle-kit, not the inline DDL

SQLite builds its tables from ~2000 lines of `sqlite.exec()` in `index.ts` and patches them forward with `migrateSchema()`. None of that is portable. Postgres and MySQL use the generated migrations in `drizzle/<dialect>/` instead — the migrator records what it applied, so it is safe to run on every start, including against a database another instance already migrated.

`drizzle/` was **not** in the production image. Added to the Dockerfile; without it the remote path would have failed at first boot with a missing-folder error, and only for people using the new feature.

### `getSqlite()` now says what is actually wrong

```
No SQLite handle: DATABASE_DIALECT is "postgres". This caller needs a
synchronous query, which only SQLite offers — give it an async path instead.
```

The old message was "SQLite not initialized. Ensure initializeDatabase() is called" — which on Postgres is both false and points at the wrong fix. There is one such caller left (`getCurrentSettingValue`), and it already checks the dialect and reads the cache instead.

`db` and `getDb()` are now typed `PortableDatabase` rather than `ReturnType<typeof drizzle<typeof schema>>`, so the handle stops claiming to be better-sqlite3-specific.

### State

**185 files, 1302 passed, 1 skipped.** Lint 0 errors, build clean, format clean.

### What this still does not do

I want to be exact about what "wired up" means here, because it is easy to over-read.

Setting `DATABASE_DIALECT=postgres` will now **connect and migrate.** It will not yet run the application correctly: the ~34 call sites that consume rows from `.returning()` are still unmigrated, and on MySQL those will throw. Postgres is closer — it has RETURNING — but has not been exercised end to end.

**This PR stays a draft until that work is done and I have actually booted the thing against both engines.** Connecting is not the same as working, and I would rather say so than let the commit message imply otherwise.

ZacharyZcR:
## 106 of 175 RETURNING sites are gone

### The framing in my earlier comment was wrong, and it mattered

I said the 92-ish majority "need no change to the query" — read the result differently and MySQL is fine.

**That was wrong, and it would have failed at runtime.** drizzle's mysql-core does not expose `.returning()` *at all*. It is not a clause MySQL rejects; it is a method that does not exist. `.returning()` on MySQL is a `TypeError`, not a bad query. Reading the result differently does not help if the statement never runs.

So the query does have to change. The good news is that it changes by getting **smaller**.

### These call sites never wanted the rows

The pattern, 106 times:

```ts
const deleted = await this.context.drizzle
  .delete(alertRules)
  .where(and(eq(alertRules.id, id), eq(alertRules.userId, userId)))
  .returning({ id: alertRules.id });

if (deleted.length === 0) return false;
```

It asks the database to send back a column it then throws away, to learn a number the driver already reported. Now:

```ts
const result = await this.context.drizzle
  .delete(alertRules)
  .where(and(eq(alertRules.id, id), eq(alertRules.userId, userId)));

if (rowsAffected(result) === 0) return false;
```

Portable, and one less thing on the wire. **This is not a MySQL concession — it is what the code should have said on SQLite too.**

### I measured the shapes instead of guessing them

Every driver reports the count. None agree on the name:

| driver | write result |
|---|---|
| better-sqlite3 | `{ changes: 1, lastInsertRowid: 7 }` |
| node-postgres | `{ rowCount: 3, rows: [], command }` |
| mysql2 | `[{ affectedRows, insertId }, fields]` |

The SQLite column is from actually running it against better-sqlite3, not from reading types. That run also turned up something worth knowing:

```
delete      : {"changes":1,"lastInsertRowid":2}
delete none : {"changes":0,"lastInsertRowid":2}
```

**`lastInsertRowid` survives the statement that set it.** After a delete it still reports the last insert's id. So `insertedId()` cannot be a general "id from this write" helper — passing it an update or delete result yields a stale id, not null. It is documented as insert-only, in bold, because the failure is silent.

MySQL's `insertId` behaves the same way. This is the kind of thing that turns into a bug report six months out with no obvious cause.

### Scope of the change

**37 files, 106 sites.** Applied by script and reviewed as a diff, not hand-edited 106 times.

`.returning()` is gone from every one, so **these paths are now MySQL-safe.** No behaviour change on SQLite: `changes` and the old `.length` are the same number, which is what the existing 1305 tests are checking.

Six variables ended up named `xRowsResult` where `result` was taken. Renamed to `xResult` — "rows" is a lie now, there are no rows.

### State

**185 files, 1305 passed, 1 skipped.** Lint 0 errors, build clean, format clean.

**69 sites left**, all of which genuinely consume returned rows:
- ~26 destructure a single row (`const [created] = await ...`) — mostly to read an autoincrement id
- ~39 read fields off the returned rows
- 4 span multiple lines and need reading individually

Those are the ones needing a read inside a transaction on MySQL. Next, and they will be hand-written, not scripted.

ZacharyZcR:
## Update and delete now work without RETURNING

29 more sites migrated. **50 left, all inserts** (plus 3 odd shapes noted below).

### The ordering is the whole thing

`updateReturning` / `deleteReturning` pick one statement or two:

```ts
if (supportsReturning(context.dialect)) {
  return db.delete(table).where(where).returning();
}
return db.transaction(async (tx) => {
  const rows = await tx.select().from(table).where(where);
  await tx.delete(table).where(where);
  return rows;
});
```

**Update reads after the write; delete reads before it.** Backwards either way and you get rows describing the wrong state — no error, no exception, just wrong data returned to the caller.

Both statements are in a transaction. Not decoration: without it a concurrent write between them makes the returned rows describe a state that never existed, and with a connection pool the second statement need not even reach the same connection.

Since no CI here runs MySQL, that ordering is asserted against a recording stub:

```ts
expect(calls).toEqual(["begin", "update", "select", "commit"]);
expect(calls).toEqual(["begin", "select", "delete", "commit"]);
```

Not a substitute for a real engine. But it fails loudly if someone later "tidies" the order.

### A bug I introduced, and why it is worth writing down

The helper returns whole rows. **11 call sites had been projecting** — `.returning({ syncId: vaultProfiles.syncId })` — and returning that projection straight out as a declared `Promise<{ syncId: string | null } | null>`.

Dropping the projection meant those methods started returning **every column of the row**, including `password`-adjacent fields on `sshCredentials`, to callers whose type said they were getting one field.

TypeScript did not catch it. A wider object is assignable to a narrower type, and excess-property checking does not apply to a variable. **The type signature was correct and the code was wrong.**

Tests caught 5 of the 11. The other 6 had no coverage — I found them by diffing every removed `.returning({...})` against what replaced it, which is why I grep for those rather than trusting the test suite. Each now picks its field explicitly:

```ts
return rows[0] ? { syncId: rows[0].syncId } : null;
```

**This is the argument against a `columns` parameter on the helper.** I considered one; it would have preserved the projections automatically and hidden the fact that 11 methods were quietly over-fetching. The explicit pick is more typing and says what it does.

### Two sites that should never have used a helper

`renameFolder` was projecting `{ id: hosts.id }` on two updates, then reading only `.length`. Rewritten as a plain update plus `rowsAffected()`. My codemod had skipped them because the result feeds a field literally named `updatedHosts:` — the "is this only used as `.length`" check saw `updatedHosts:` and bailed. Correctly conservative, and I would rather it bail than guess.

### One limitation, stated plainly

On MySQL the update path re-reads using the same `where`. **If an update changes a column that `where` tests, the read finds nothing and returns `[]` where SQLite returns the row.** Every current caller filters on an id it does not modify. It is documented in bold at the top of the helper because the failure is silent and the next person adding a caller will not think about it.

### State

**186 files, 1311 passed, 1 skipped.** Lint 0 errors, build clean, format clean.

Remaining: **47 inserts** — those need the autoincrement id, which is the one case where Postgres also cannot help without RETURNING, so it is a genuine three-way split. Plus 3 shapes that resist scripting (a ternary, a nested transaction, a multi-line chain) and will be done by hand.

ZacharyZcR:
## All 175 RETURNING sites are done

Every remaining `.returning()` in the repository layer now sits inside a branch that has already checked the dialect. Four of them, in two files, both `if (this.context.dialect === "sqlite")`.

### Insert was the only genuinely three-way case

Update and delete collapse to "one statement or two". Insert does not: **without RETURNING there is no id to read the row back by**, and that is true of Postgres too. The saving grace is that both SQLite and Postgres *have* RETURNING, so it collapses to two branches again.

On MySQL the key comes from one of two places, and which one is a property of the table:

- **the caller supplied it** — `users`, `sessions`, `apiKeys` and four others are keyed by a text id generated in application code
- **the engine assigned it** — 25 tables with an autoincrement `id`, reported as `insertId`

`insertReturning` is constrained to `SQLiteTable & { id: SQLiteColumn }`. That was not decoration — **the codemod tried to convert `user_preferences` and the compiler stopped it**, because that table is keyed by `userId` and has no `id` column at all. It would have inserted fine and then read back nothing. A runtime `undefined` in a signature promising a record, on a table that stores user settings.

That site got `insertReturningWhere`, which takes the read-back condition explicitly.

### better-sqlite3 will not run an async transaction

Three sites used synchronous transactions with `.all()`. I checked what actually happens rather than reasoning from the types:

```
async tx FAILED -> Transaction function cannot return a promise
rows after: []
```

**It refuses, and the write does not happen.** That is the one mercy here — it fails loudly instead of committing early and silently dropping the body.

But it means no single body serves both. The transaction has to be written twice, sync for SQLite and async for everything else. That is not a style preference and I do not want it read as one, so it is commented at both sites.

The two user methods were **already duplicates of each other** — `createFirstLocalUser` and `createFirstSsoUser` differed only in how they computed `isAdmin`. Folding them into one private method with a `build(isFirstUser)` callback means the branch is written once, not four times. **Net fewer lines than before, despite gaining a dialect branch.**

Worth being clear on why that transaction exists at all: it makes "is the table empty" and "insert" atomic. Two people signing up simultaneously would otherwise both see an empty table and **both become admin**.

The CA repository has the same shape for a different reason: the private key is encrypted with the row's own id as context, which does not exist until the row does. Insert blank, encrypt, update — and the blank key must never be observable.

### Where this leaves the branch

| | |
|---|---|
| RETURNING call sites migrated | 175 |
| Repository files touched | 38 |
| Sync-only driver calls left in repositories | 1 (`getCurrentSettingValue`, already dialect-aware) |

**186 files, 1311 passed, 1 skipped.** Lint 0 errors, build clean, format clean.

### Still a draft, and here is the honest reason

The repository layer is portable now. **It has not been run against Postgres or MySQL.** Every claim above rests on generated schema, driver shapes I measured, and a test suite that only exercises SQLite.

What I would expect to still be wrong: `sql\`CURRENT_TIMESTAMP\`` and `substr()` in the folder-rename expression are SQLite spellings; `datetime(...)` arithmetic in the retention queries likewise. Those are query text, not call shape, so this pass would not have caught them.

Standing up both engines and booting against them is the next step, and nothing here should merge before that happens.

ZacharyZcR:
## Run against real Postgres and real MySQL

I said this should not merge until it had actually been booted against both engines. It has now.

- **PostgreSQL 16.14**
- **MariaDB 10.11.14** (MySQL wire protocol)

### Migrations

`drizzle/postgres` → **52 tables**. `drizzle/mysql` → **53** (the extra is drizzle's own migration journal). Both applied clean on an empty database, first try.

### Startup

All three dialects reach a working `getDb()`:

```
[INFO] Connecting to postgres database [op:db_init]
[INFO] postgres database ready [op:db_init_complete]
BOOT OK  dialect=postgres  users=4

[INFO] Connecting to mysql database [op:db_init]
[INFO] mysql database ready [op:db_init_complete]
BOOT OK  dialect=mysql  users=3

[INFO] Initializing SQLite database [op:db_init]
BOOT OK  dialect=sqlite  users=0
```

The mismatch guard fires as intended:

```
DATABASE_URL is a "mysql://" URL but DATABASE_DIALECT is "postgres".
Expected one of postgres://, postgresql://.
```

And `getSqlite()` on a remote dialect says the useful thing rather than "not initialized".

Worth noting: the `DATA_DIR` guard only runs on the SQLite path. Postgres and MySQL deployments have no data directory to misconfigure, and correctly never reach it.

### Repositories, driven for real

Not the test suite — the actual repository classes, against actual servers. **20/20 operations on MySQL, 17/17 on Postgres**, covering `insertReturning`, `updateReturning`, `deleteReturning`, `insertReturningWhere`, the async transaction branch, and the `substr()`/`CASE WHEN` folder rename.

### Then I checked the values, not just the absence of exceptions

"It did not throw" is the wrong bar for this change. The MySQL paths are two statements in a transaction, and getting the **order** wrong returns plausible, wrong data silently. So:

| assertion | postgres | mysql |
|---|---|---|
| update returns the NEW value | ok | ok |
| delete returns pre-delete rows | ok | ok |
| rows really gone afterwards | ok | ok |
| autoincrement id came back | ok | ok |
| database-default `createdAt` came back | ok | ok |
| boolean `true` survives round-trip | ok | ok |

That last row deserves a note, because **it works by luck.** Repositories still import the SQLite schema, where booleans are `integer(mode: "boolean")`. drizzle's SQLite mapper is:

```js
mapFromDriverValue(value) { return Number(value) === 1; }
mapToDriverValue(value)   { return value ? 1 : 0; }
```

Writing sends `1`, which Postgres and MySQL both accept for a boolean column. Reading gets back JS `true`, and `Number(true) === 1` is true. **Both directions survive by coincidence of `Number()` coercion**, not by design. It is correct today; I would not want anyone to think it was engineered.

Which points at the real remaining gap: **44 repository files statically import `db/schema.js`, and `schemaFor()` has no callers.** Every engine is currently driven through SQLite table definitions. It works — drizzle largely reads names and columns off them — but it is not the intended design and it is why `PortableDatabase` is still an approximation.

### Where the branch stands

**186 test files, 1311 passed, 1 skipped.** Lint 0 errors, build clean, format clean. SQLite behaviour unchanged.

The claims in my earlier comments were reasoning from types and generated SQL. They now have a machine behind them. Two things I predicted would break did not — `substr()` and the timestamp comparisons — because `sqlTimestampDaysAgo` already computes cutoffs in JS instead of calling `datetime()`.

Still draft. What I would want before taking it off draft: wiring `schemaFor()` through so each dialect uses its own table definitions rather than relying on the coincidence above, and a CI job that runs the suite against all three.

ZacharyZcR:
## The big refactor I said was next turned out to be unnecessary

My last comment ended with: wire `schemaFor()` through so each dialect uses its own table definitions. That would have touched all 44 repository files.

I looked into what it would actually buy before starting. **Nothing.** Here is the reasoning, because the conclusion is more useful than the diff.

### What a table object is for

drizzle needs exactly two things from one to build a query: the **identifiers** to interpolate, and the **encoders** that turn JS values into driver values. It does not consult the column type for anything else — that only matters when generating DDL.

The schema has four column shapes, and that is the entire surface:

| shape | count | encoder |
|---|---|---|
| `text` | 386 | identity |
| `integer` | 114 | identity |
| `integer({ mode: "boolean" })` | 50 | `1/0` out, `Number(v) === 1` back |
| `real` | 5 | identity |

Three of the four are identity functions. The fourth writes `1`, which Postgres and MySQL both accept for a boolean column, and reads back through `Number(v) === 1`, which holds for JS `true` as well as for `1`.

So the sqlite definitions are **correct on all three engines**, and the generated schemas exist for exactly one purpose: drizzle-kit reads them to emit DDL.

I called the boolean case "luck" last time. Having traced why it holds in both directions I would now call it narrow but sound — and either way, the fix for a fragile invariant is to assert it, not to restructure 44 files around it.

### What I removed instead

- **`schemaFor()` — deleted.** I added it earlier in this branch. It had no callers, and its existence advertised a runtime dialect switch that does not exist and is not needed. Dead code that lies about the design is worse than dead code.
- **`connect.ts` no longer loads the generated schema.** It was passing `{ schema }` to `drizzle()`, which only feeds the relational query API (`db.query.*`). Nothing in this codebase uses that. It was loading a 52-table module per connection for an API with no callers.

Net: **fewer moving parts than my previous comment proposed, and one less thing implying a design that was never there.**

### And what I added, so this stops being a matter of opinion

`scripts/verify-dialects.mjs`, wired up as `npm run verify:dialect <url>`. It drives the real repository classes against a live server and asserts values, not the absence of exceptions:

```
verifying mysql at mysql://termix:***@127.0.0.1:3306/termix_test

  ok    insert returns the stored row
  ok    boolean true survives the round trip
  ok    update returns the new value
  ok    autoincrement id came back
  ok    database-assigned createdAt came back
  ok    delete reports the rows it removed
  ok    and they are actually gone
  ok    host really deleted

mysql: all checks passed
```

Identical output for Postgres. It exits non-zero on failure, so it can be a CI job the moment someone stands up service containers.

The two ordering checks are the ones that matter. `update` must report state **after** the write and `delete` **before** it; reverse either and MySQL returns plausible, wrong data with no error anywhere. Those are now asserted against a real server instead of a stub.

The generated schema headers say what they are for, so the next person does not repeat my mistake:

```
// DDL source for drizzle-kit. NOT imported to run queries — repositories use
// schema.ts on every dialect. See the generator header for why that is correct.
```

### State

**186 test files, 1311 passed, 1 skipped.** Lint 0 errors, build clean, format clean, schema drift green. Live verification passing on PostgreSQL 16.14 and MariaDB 10.11.

### Ready to come out of draft

The two conditions I set last time were the `schemaFor()` wiring and a repeatable cross-engine check. The first is answered by not needing it; the second is committed and passing.

What is still missing is CI actually running it — that needs Postgres and MySQL service containers in the workflow, which I would rather do as its own PR than bundle here.

ZacharyZcR:
## CI is green, and it caught a bug on its first run

I changed my mind about splitting the CI job into its own PR. A capability that nothing guards decays, and between merging this and merging that there would have been a window where nothing checked either engine. The job belongs with the thing it protects.

That decision paid for itself immediately.

### My local testing was wrong about MySQL

I verified against **MariaDB 10.11**, because that is what installs cleanly here. CI runs **MySQL 8**. First run:

```
Error: Invalid default value for 'fired_at'
  code: 'ER_INVALID_DEFAULT', errno: 1067
  sql: CREATE TABLE `alert_firings` (
         `fired_at` text NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

**MySQL only accepts a bare `DEFAULT CURRENT_TIMESTAMP` on a DATETIME or TIMESTAMP column.** Timestamps here are text (see `sql-timestamp.ts` — they are compared lexicographically so retention needs no date functions). Since 8.0.13 an expression default works on any type, but an expression must be parenthesised: `DEFAULT (CURRENT_TIMESTAMP)`.

**MariaDB accepts the bare form.** So every local run passed while the generated DDL was invalid for the engine the feature is named after. I would not have found this by reading, and I had already reported MySQL as working — that report was wrong, and this is the correction.

The generator now emits the parenthesised form for MySQL only, across all 71 timestamp columns; Postgres is untouched. `drizzle/mysql` was regenerated.

Worth noting what the failure hid: it aborted on `alert_firings`, the first table. **The other 50-odd tables were never attempted.** A green MariaDB run told me nothing about them.

### Second run

```
verifying postgres  →  8/8   postgres: all checks passed
verifying mysql     →  8/8   mysql: all checks passed
```

Both jobs pass. `lint-and-build` passes.

### One more gap closed

`pr-check.yml` ran `npx eslint .` rather than `npm run lint`. The npm script also runs `generate-dialect-schema.cjs --check`, which catches a `schema.ts` edit that forgot to regenerate the dialect copies. **That check has never run in CI** — I added it to the script earlier in this branch and did not notice the workflow bypassed it. Now switched to `npm run lint`.

### What the job does

Service containers for `postgres:16` and `mysql:8`. Each run applies the migrations to an **empty** database through the same `runRemoteMigrations()` the application calls at startup, so a migration that does not apply cleanly fails the build — which is precisely how the bug above surfaced.

Locally: `npm run verify:dialect -- <url>`.

### State

**186 test files, 1311 passed, 1 skipped.** Lint 0 errors, build clean, format clean, schema drift green. CI green on PostgreSQL 16 and MySQL 8.

Out of draft. Ready for review.

The honest summary of this PR: SQLite behaviour is unchanged and covered by the existing suite; Postgres and MySQL now work and are covered by a job that has already proven it catches things I could not catch by hand.

ZacharyZcR:
## Two things that were documented rather than enforced

I asked myself what could still go wrong with this once it is merged and I stop looking at it. Two answers, both now fixed. A third I measured and decided against.

### 1. Nothing stopped call site number 176

175 sites were migrated off `.returning()`. **Nothing prevented the next one from being written the old way** — and on MySQL that is a `TypeError`, on a path no test in this repo exercises. It would reach a user before it reached a test.

Now an eslint rule:

```
error  MySQL has no RETURNING. Use insertReturning/updateReturning/deleteReturning
       from ./returning.js, or rowsAffected() if you only need a count.
```

It fires at the point of writing, in the editor, in the CI job that was already running.

Four legitimate uses remain, all inside branches that have already checked the dialect. Each carries a disable comment naming the reason rather than being excluded wholesale by filename:

```ts
/* eslint-disable no-restricted-syntax -- sqlite-only branch: the dialect is
   checked directly above, and better-sqlite3 rejects an async transaction
   callback, so this cannot use the shared helpers. */
```

I verified the rule actually catches a violation by adding one, watching it fail, and reverting.

### 2. The silent trap is now a loud one

My earlier comment described this in bold and left it at that:

> On MySQL the update path re-reads using the same `where`. If the update changes a column that `where` tests, the read finds nothing and you get `[]`.

**A comment does not stop anyone.** The condition is detectable — rows changed, none readable back — so it now throws:

```
updateReturning wrote 3 row(s) but could not read them back: the update changed
a column the where clause filters on. Read the rows first, or filter on a column
the update leaves alone.
```

The distinction that matters, and which the tests pin: **wrote-but-cannot-read throws; genuinely-matched-nothing still returns `[]`.** Conflating those would be its own bug.

Same treatment for both insert paths — a row that cannot be read back after insertion throws instead of returning `[]` to a caller whose signature promises a record.

This is the change I am most confident about. Every one of these would have been a silent wrong answer on MySQL only, which is the hardest kind to ever find.

### 3. What I measured and did not do

The dialect job exercises **3 of 43 repositories**. The obvious fix is to run the existing suite against Postgres. I checked whether that was cheap:

- 45 repository test files
- **41 reach for SQLite-specific driver access** (`.raw`, `.exec`, `.prepare`)
- **0 are engine-agnostic as written**

So it means rewriting the fixture and 41 test files. That is a real improvement and a separate PR — bundling it here would bury this diff. Recording the number so the gap is a known quantity rather than an impression.

### State

**186 test files, 1314 passed, 1 skipped.** Lint 0 errors, build clean, format clean. CI green on PostgreSQL 16 and MySQL 8.

### On upgrade safety, since it came up

Verified rather than asserted:

- This PR changes **no** `schema.ts`, **no** CREATE/ALTER TABLE, and none of the encryption, key, save-trigger or migration code. No data migration exists to go wrong.
- I diffed the repository layer against the merge base behaviourally: 13 scenarios covering the rewritten projections that previously had no test, run against identical databases through both versions. **No differences.** The differ was negative-controlled by injecting a difference and confirming it was caught.
- A database populated by the base build, then opened by this build: **53 tables identical row for row** once timestamps are masked.

Unrelated, found along the way and left alone as out of scope: with `DB_FILE_ENCRYPTION=false` the startup path always creates a fresh in-memory database and never loads `db.sqlite`. Present in the base too, not introduced here. Worth its own issue.

ZacharyZcR:
## The repository suite now runs on all three engines

3 of 43 repositories were covered against a real Postgres or MySQL. Now the whole suite runs on each.

```
sqlite    45 files, 241 tests
postgres  44 files, 223 passed, 18 skipped
mysql     44 files, 223 passed, 18 skipped
```

Both remote runs are CI steps. `TEST_DIALECT` + `TEST_DATABASE_URL` do it locally; unset, `npm test` behaves exactly as before and needs no server.

### What the tests were actually running against

Each file hand-wrote the DDL for the tables it touched. `users` was declared with five columns where the real one has thirty. One file created a table named `hosts`; the real table is `ssh_data`. **No foreign keys anywhere.**

So the suite was green against a schema that does not exist, and fixtures referenced rows that could not exist — a `host_access` row pointing at host 44 when only host 42 was ever inserted.

The fixture now builds the **real** schema, from a generated migration, with foreign keys enforced throughout. That produced 109 failures, every one of them a fixture asserting something impossible. All fixed; **41 of the 45 files needed edits**, and 120 blocks of hand-written DDL are gone.

### Five defects this found

**1. Four composite unique constraints existed only in production.**

`user_roles(user_id, role_id)`, `shared_host_secrets(host_access_id, target_user_id, protocol)`, `opkssh_tokens(user_id, host_id)`, `vault_tokens(user_id, profile_id)` — declared inline in the CREATE TABLE in `index.ts`, never in `schema.ts`. **All three generated schemas were missing them.** On Postgres and MySQL those tables would have permitted duplicate rows SQLite forbids, and the upserts had no constraint to conflict on.

**2. Three writes read better-sqlite3 fields directly.**

```ts
return Number(result.lastInsertRowid);
```

`undefined` on the other drivers, so `Number(undefined)` is `NaN`, which reaches Postgres as the string `"NaN"` and fails an integer column. My earlier codemod only searched for `.returning()` and never saw these.

**3. A COUNT annotation is an assertion, not a conversion.**

Postgres returns COUNT as bigint, which node-postgres hands back as a **string** so large values survive. `count < max` was comparing a string to a number. `countValue()` now does the conversion the annotation only claimed.

**4. Export order depended on the engine.**

`ORDER BY folder ASC` where folder is nullable. **NULLs sort first on SQLite and MySQL, last on Postgres.** Fixed with a coalesce so the order is the same everywhere.

**5. `onConflictDoUpdate` does not exist on MySQL.**

Same shape as the RETURNING problem: MySQL spells it `ON DUPLICATE KEY UPDATE` and names no columns, so drizzle's mysql-core has no such method and the call is a TypeError. Two upserts went through a new `upsert()` helper.

Each of 1, 2 and 5 now has an eslint rule, so the next occurrence fails at the point of writing rather than on the one engine no test could reach.

### CI caught one more that my machine could not

I develop against **MariaDB 10.11**; CI runs **MySQL 8**. After adding the unique constraint on `shared_host_secrets`, CI failed:

```
BLOB/TEXT column 'protocol' used in key specification without a key length
```

The generator already widens key-bearing text columns to `varchar(255)`, but it only looked at column-level modifiers — a **table-level** uniqueIndex was invisible to it, so `protocol` stayed TEXT. MariaDB accepted the migration; MySQL 8 did not.

That is twice now that the difference between my machine and CI has been the thing that found the bug. Worth stating plainly: **MariaDB is not a stand-in for MySQL**, and I should not have treated it as one.

### The 18 skipped

Tests that reach past drizzle into the SQLite driver to assert on stored bytes — that encryption really happened at rest. Not portable by nature, so they skip rather than fail, and **all 18 still run on SQLite**. What is lost on the other engines is the byte-level assertion, not the behaviour.

### Fixture notes

- migrations run once per worker, not once per fixture
- `TRUNCATE` on all 53 tables costs ~2s per test on MySQL, so it asks which tables hold rows first
- Postgres sequences and MySQL auto_increment counters are moved past ids the fixture inserted by hand; SQLite needs no help
- booleans in seed SQL are rewritten to TRUE/FALSE from the schema, since SQLite stores 0/1 and Postgres rejects an integer in a boolean column

### State

**186 test files, 1314 passed, 1 skipped** on SQLite. Lint 0 errors, build clean, format clean, schema drift green. CI green: PostgreSQL 16 and MySQL 8, both the smoke checks and the full repository suite.

ZacharyZcR:
## A gap my own verification had missed

Checking whether anything outside the repository layer still reaches for the SQLite driver, I found four call sites that do — and **two of them are on the unconditional startup path**. On Postgres the application would have failed to boot:

```
legacy-share-cleanup: No SQLite handle: DATABASE_DIALECT is "postgres".
shared-host-secrets:  No SQLite handle: DATABASE_DIALECT is "postgres".
```

### Why I did not catch this earlier

My boot test called `initializeDatabase()` and stopped there. The real sequence in `starter.ts` runs several data migrations afterwards, and those are where the synchronous driver is used. **"BOOT OK" was measuring less than it implied**, and I reported it as though it covered startup. It did not.

### What the four are, and what they do now

Three are SQLite-era data migrations: legacy shared-credential column cleanup, the shared-host-secrets rebuild, and the per-user field-encryption backfill. They repair shapes that **only a pre-existing SQLite database can have** — a Postgres or MySQL database is created by the drizzle migrations, which have never emitted any of them. So each returns early on other engines, for the reason stated at the call site rather than a bare dialect check.

The fourth is different and I did not want to paper over it.

**Importing a backup relies on `PRAGMA foreign_keys = OFF`** because the restore writes tables in an order that is not dependency-safe. There is no equivalent available here: Postgres needs superuser to disable triggers, and MySQL's `SET FOREIGN_KEY_CHECKS = 0` is per-connection, which a pool does not guarantee.

The choices were to run the import with constraints enforced — which fails partway through and leaves a **half-restored database** — or to refuse. It refuses:

```
Importing a backup is only supported on SQLite; this deployment uses postgres.
Restore into the database directly with its own tooling instead.
```

A feature that says no is better than one that corrupts. This is now a documented limit rather than a surprise.

### What was already fine

Worth recording, since I checked: the 15 `.prepare()` calls in `database.ts` all operate on the **export/import file itself**, which is a standalone SQLite artifact regardless of backend. And `migrateAuditRetention` is inside `initializeCompleteDatabase()`, which the remote path returns before reaching. Neither needed changing.

`getCurrentRepositorySqlite()` now has four callers and every one sits behind a dialect check.

### State

**186 test files, 1314 passed, 1 skipped** on SQLite; 44 files / 223 passed / 18 skipped on each of Postgres and MySQL. Lint 0 errors, build clean, format clean. CI green on PostgreSQL 16 and MySQL 8.

`docs/database-backends.md` records the backup-import limit and the skipped migrations under Known limits.

ZacharyZcR:
## Silent wrong data on MySQL, and how it hid from the suite

Sweeping for dialect differences that produce **wrong answers rather than errors**, one turned up.

### `||` is not concatenation on MySQL

`renameFolder` built child folder paths with SQLite's concatenation operator:

```sql
CASE WHEN folder = 'prod' THEN 'ops' ELSE 'ops / ' || substr(folder, 8) END
```

Measured, not assumed:

```
mysql     SELECT 'ab' || 'cd'  ->  0
postgres  SELECT 'ab' || 'cd'  ->  abcd
```

MySQL reads `||` as logical OR unless the server runs with `PIPES_AS_CONCAT`. So renaming a folder would have rewritten every child path to **`0`** — no error, no exception, just folder names replaced with a zero.

Now `CONCAT(...)`, which SQLite (3.44+), Postgres and MySQL all agree on.

Two details that only showed up against real servers:

- Postgres rejected the parameterised form with **42P18, "could not determine data type of parameter"** — `CONCAT` is variadic, so there is no positional type to infer from. The prefix needs an explicit `CAST`.
- The cast target is not the same word: MySQL has no `text` cast and wants `char`.

### The suite had a test for this and it was skipped

The assertion existed and was correct:

```ts
expect(...).toEqual([{ folder: "ops" }, { folder: "ops / api" }]);
```

But it read the rows through the raw SQLite handle, so it carried `itSqliteOnly` and **never ran on the engine where the bug was**. Rewritten to use the fixture's portable query; it now fails on MySQL before the fix and passes on all three after.

**That is the real lesson from this one.** The 18 SQLite-only tests are not all genuinely SQLite-only — some were only reaching for the raw driver out of convenience, and each one is a place a portability bug can hide. I have not audited the remaining 17 for the same mistake.

### Also swept

- No other `||` in any SQL template. An eslint rule now rejects one inside a `sql` tag, verified by adding a violation and watching it fail.
- No `strftime`, `julianday`, `IFNULL`, `GROUP_CONCAT` or `RANDOM()` anywhere in the repository layer.
- **`LIKE` is case-insensitive on SQLite and case-sensitive on Postgres.** Four call sites match folder prefixes and settings keys, so renaming `prod` on SQLite also catches `PROD / api` and on Postgres does not. Postgres is arguably the better behaviour of the two, so nothing was changed to force agreement — that would alter what existing SQLite deployments do. Recorded under Known limits.

### State

**186 test files, 1314 passed, 1 skipped** on SQLite; 44 files / 223 passed / 18 skipped on each of Postgres and MySQL. Lint 0 errors, build clean, format clean. CI green on PostgreSQL 16 and MySQL 8.

Four eslint rules now cover the classes found in this branch: bare `.returning()`, `lastInsertRowid`/`changes`, `onConflictDoUpdate`, and `||` inside a `sql` tag.

ZacharyZcR:
## The skipped tests were hiding a second bug

Auditing the 18 SQLite-only tests: **none of them were actually SQLite-only.**

Every one used plain SQL — `SELECT id FROM ssh_data ORDER BY id`, `SELECT password FROM ssh_credentials WHERE id = ?`. What tied them to SQLite was the `.prepare()` **API**, not the SQL. They reached for the raw driver out of convenience, and carried a skip marker as the price.

All 18 now run everywhere:

```
sqlite    45 files, 241 tests
postgres  45 files, 241 tests
mysql     45 files, 241 tests
```

**Zero skipped**, on all three.

### What that immediately caught

`createEncryptedForUser` in the CA repository, in code I wrote earlier in this branch:

```ts
const written = await tx.insert(termixIdentityCa).values(draft);
const id = supportsReturning(this.context.dialect)
  ? (written as unknown as { id: number }[])[0].id   // no .returning()
  : insertedId(written);
```

The RETURNING branch never attached `.returning()`, so it indexed into a write result that has no rows. `TypeError: Cannot read properties of undefined` on Postgres — and **no test could reach it**, because the only test covering that method was skipped.

The cast is what let it through: `as unknown as { id: number }[]` told the compiler to stop checking exactly where checking would have helped.

### A process note

My first attempt at this was a regex sweep over all five files at once. It produced `repo.await adapter!.run(...)`, ate the closing brace of unrelated template literals like `` `action_${i}` ``, and left the suite unable to parse. I reverted the whole thing and did it file by file, verifying after each.

Worth recording because the earlier codemods in this branch **did** work, and it would be easy to conclude the technique is always fine. It is not: those transformed one well-delimited construct, this one had to rewrite parameter binding inside string literals.

### Net effect

**8 files changed, 438 insertions, 537 deletions** — 99 lines fewer. The fixture's `raw` accessor, `itSqliteOnly` and `describeSqliteOnly` are gone, along with the per-file `get sqlite()` plumbing, because nothing needs them now.

One fixture fix was needed: `run()` used drizzle's `all()`, and better-sqlite3 refuses that on a statement returning no rows.

### State

**186 test files, 1314 passed, 1 skipped** on SQLite; **241 of 241 on each of Postgres and MySQL**. Lint 0 errors, build clean, format clean. CI green on PostgreSQL 16 and MySQL 8.

The claim "the repository layer is covered on all three engines" is now literally true rather than true of 92% of it.

ZacharyZcR:
## The settings cache contradicted the reason this exists

The doc says Postgres and MySQL are here for deployments that need **more than one process to reach the data — multiple replicas**. The settings cache is a module-level `Map`, updated by whichever process performed the write.

So on exactly the deployment this feature is for: replica A changes a setting, and **replica B serves the old value until it restarts.** Not for a while — indefinitely.

That is fine on SQLite, where there is only ever one process. It is not fine here, and I had documented the multi-replica story without noticing it.

### Why not just make the read async

Because `getCurrentSettingValue` has 27 synchronous callers — during startup, inside request handlers, from the guacd bootstrap. That is the whole reason the cache exists. A synchronous read cannot go back to the database, so it cannot be made correct on demand.

What it can be is **bounded**. Each replica re-reads the settings table every 30 seconds, `SETTINGS_CACHE_REFRESH_SECONDS` to change it, `0` to turn it off:

- the replica that made the change sees it at once, through the existing write path
- the others see it within the interval

This is not consistency and the doc does not call it that. It is the difference between a setting that takes effect on the next tick and one that takes effect at the next restart.

### The failure case is the interesting one

A refresh that throws leaves the previous values in place:

```ts
void primeCurrentSettingsCache().catch(() => {
  // A failed refresh leaves the previous values in place, which is the
  // right outcome: a transient database blip should not blank the cache.
});
```

Every caller reads a missing setting as "use the default". So a cache emptied by a momentary connection reset would silently revert configuration across the deployment — registration re-enabled, guacd URL lost — until the next successful refresh. Pinned by a test.

The timer is `unref()`d so it does not hold the process open.

### State

**187 test files, 1317 passed, 1 skipped** on SQLite; **244 of 244 on each of Postgres and MySQL**. Lint 0 errors, build clean, format clean. CI green on PostgreSQL 16 and MySQL 8.

PR #1138, opened 2026-07-29
fix: compare sync cursors independently of timestamp layout

kacperpietrzyk:
## Summary

- compare sync cursors on a normalized `YYYY-MM-DD HH:MM:SS` form instead of trusting the stored layout
- apply it to both the row pull (`GET /sync/:entityType`) and the tombstone pull (`listSince`)
- cover rows written by `CURRENT_TIMESTAMP` and by `toISOString()`, plus the cursor's own second

Incremental sync returned nothing at all after its first pass. `gt(table.updatedAt, since)` runs on
a TEXT column whose values default to `CURRENT_TIMESTAMP` (`2026-07-29 10:11:21`), while the desktop
engine sends `new Date().toISOString()` (`2026-07-29T10:06:55.172Z`). Text comparison is decided at
position 10, where `' '` (0x20) sorts below `'T'` (0x54), so the predicate answers on layout rather
than on time — and is false for every `CURRENT_TIMESTAMP` row however new it is:

```
GET /sync/hosts                                 -> 3 rows (newest 2026-07-29 10:11:21)
GET /sync/hosts?since=2026-07-29T09:00:00.000Z  -> 0 rows    <- the format the client sends
GET /sync/hosts?since=2026-07-29 09:00:00       -> 3 rows    <- the format the column stores
```

The engine only sends a cursor from the second pass onward, so pass 1 synced everything and passes
2..n pulled zero rows and zero tombstones while reporting `lastError: null`. Edits and deletions
stopped propagating in both directions, silently.

Not a regression from #1092 — this code is unchanged since #1085. It was *masked*: before the
reference fix the loop threw before persisting state, so the cursor never advanced past `null` and
every cycle was a full sync. Now that a pass completes, the latent defect is reachable, and it will
reach users as 2.6.1 rather than as a visible change.

## Notes on the approach

- `replace`/`substr` rather than `datetime()`: the repository layer is deliberately drizzle-only so
  that no repository reaches for engine-specific SQL, and #1134 is laying Postgres/MySQL
  groundwork, so I avoided the SQLite-only date function. To be precise about what I actually
  verified: this is tested on SQLite only. `replace`/`substr` are much likelier to survive a
  dialect change than `datetime()`, but I can't claim Postgres/MySQL compatibility I haven't run.
  If the multi-dialect work needs a different expression, `sync-timestamp.ts` is the single place
  to change — and I'm happy to move it behind the engine boundary instead if you'd rather own it
  there.
- `>=` rather than `>`: normalising truncates sub-second precision, so a strict comparison would
  permanently skip rows written in the cursor's own second. Re-sending that boundary is a no-op —
  `syncEntity()` pushes only when one side is strictly newer, so rows equal on both sides do
  nothing.
- `updatedAt` is written in both layouts across the codebase (14 sites use `toISOString`, 11 use
  `CURRENT_TIMESTAMP`), so the tests assert on rows of each kind rather than assuming one.

## Validation

- `npm test -- --run src/backend/tests/database/repositories/sync-tombstone-since.test.ts` (4 passed)
- `npm test` — 181 files, 1248 passed, 1 skipped
- `npm run type-check`
- `npm run lint` (0 errors; the 44 existing warnings only)
- `npm run build`
- end to end: server image built from this branch on an isolated volume, driving the real
  `electron/remote-sync.cjs` against it with a throwaway profile. Before: pass 2 pulls nothing and a
  rename made on the server never reaches the client. After: the rename arrives on the next pass.

Deletions need #1051 as well — while the tombstone endpoint is unreachable there is nothing to
apply even once the cursor works.

Closes Termix-SSH/Support#1050

### DIFF
--- a/src/backend/database/repositories/sync-tombstone-repository.ts
+++ b/src/backend/database/repositories/sync-tombstone-repository.ts
@@ -1,5 +1,6 @@
-import { and, eq, gt } from "drizzle-orm";
+import { and, eq } from "drizzle-orm";
 import { syncTombstones } from "../db/schema.js";
+import { timestampAtOrAfter } from "../sync-timestamp.js";
 import type { DatabaseContext } from "./database-context.js";
 
 export type SyncTombstoneRecord = typeof syncTombstones.$inferSelect;
@@ -57,7 +58,8 @@ export class SyncTombstoneRepository {
       eq(syncTombstones.userId, userId),
       eq(syncTombstones.entityType, entityType),
     ];
-    if (since) conditions.push(gt(syncTombstones.deletedAt, since));
+    if (since)
+      conditions.push(timestampAtOrAfter(syncTombstones.deletedAt, since));
 
     return this.context.drizzle
       .select()
--- a/src/backend/database/routes/sync.ts
+++ b/src/backend/database/routes/sync.ts
@@ -1,6 +1,6 @@
 import type { Request, Response } from "express";
 import express from "express";
-import { and, eq, gt } from "drizzle-orm";
+import { and, eq } from "drizzle-orm";
 import {
   hosts,
   sshCredentials,
@@ -27,6 +27,7 @@ import {
   serializeSyncReferences,
   type SyncReferenceEntity,
 } from "./sync-references.js";
+import { timestampAtOrAfter } from "../sync-timestamp.js";
 
 const router = express.Router();
 const authManager = AuthManager.getInstance();
@@ -232,7 +233,9 @@ router.get(
       const context = createCurrentRepositoryContext();
       const conditions = [eq(table.userId, userId)];
       if (since && "updatedAt" in table) {
-        conditions.push(gt((table as typeof hosts).updatedAt, since));
+        conditions.push(
+          timestampAtOrAfter((table as typeof hosts).updatedAt, since),
+        );
       }
 
       const rows = await context.drizzle
--- a/src/backend/database/sync-timestamp.ts
+++ b/src/backend/database/sync-timestamp.ts
@@ -1,0 +1,35 @@
+import { sql, type SQLWrapper } from "drizzle-orm";
+
+/**
+ * Sync cursors and stored timestamps do not share a layout.
+ *
+ * `updated_at` and `deleted_at` are TEXT columns written both by
+ * `default(sql`CURRENT_TIMESTAMP`)` ("2026-07-29 10:11:21") and by
+ * `new Date().toISOString()` ("2026-07-29T10:11:21.123Z"), while the desktop
+ * sync engine always sends the ISO form as `since`. Comparing those as text is
+ * decided at position 10, where ' ' (0x20) sorts below 'T' (0x54), so the
+ * answer depends on which writer produced the row rather than on when it was
+ * written -- and `column > :isoCursor` is false for every row stored in the
+ * CURRENT_TIMESTAMP form, however new it is.
+ *
+ * Both layouts share a prefix once the separator is levelled, so comparing
+ * "YYYY-MM-DD HH:MM:SS" on both sides is layout-independent. `replace` and
+ * `substr` are used rather than `datetime()` to keep the expression portable
+ * across engines.
+ */
+export const CANONICAL_TIMESTAMP_LENGTH = 19;
+
+export function normalizeSyncTimestamp(value: string): string {
+  return value.replace("T", " ").slice(0, CANONICAL_TIMESTAMP_LENGTH);
+}
+
+/**
+ * `>=` rather than `>`: normalising truncates sub-second precision, so a strict
+ * comparison would permanently skip rows written in the same second as the
+ * cursor. Re-sending that boundary second costs nothing -- the sync engine only
+ * pushes a row when one side is strictly newer, so rows equal on both sides are
+ * a no-op.
+ */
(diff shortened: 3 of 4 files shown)


ZacharyZcR:
Same apology as on #1139: I opened #1140 covering the same two defects 4.5 hours after you filed this, without checking whether anyone was already working on them, and then merged it — turning both of your PRs into dead work. That was my error. #1140 is now reverted (#1146), #1139 is merged, and this is the one that should carry the cursor fix.

Two things happened on `dev-2.6.1` after you opened this, and one of them breaks this branch:

**1. `package.json` was invalid JSON from 2026-07-29 to today.** The version bump wrote `"version": "2.6.1,` without the closing quote, so every CI run on this branch failed before reaching any test — fixed in #1147. Any green checks you saw on this PR are from a run against an older base.

**2. #1134 changed the test harness, and that is what fails here now.** Rebased onto current `dev-2.6.1`, three of your tests fail:

```
SqliteError: table sync_tombstones already exists
SqliteError: table ssh_credentials already exists
  at TestSqliteDatabase.exec  (test-support.ts:96)
  at sync-tombstone-since.test.ts:69
```

`test-support.ts` now creates the schema itself as part of the multi-dialect work, so the hand-written `CREATE TABLE` blocks in `sync-tombstone-since.test.ts` collide with tables that already exist. Dropping those blocks and seeding rows against the tables the harness provides should be all it takes — the assertions themselves look right to me and are testing exactly the correct thing.

For what it is worth, I reached the same diagnosis independently and normalised the cursor server-side in `normalizeSince()`. One case worth keeping in mind if you touch that path: a value that is *already* in `YYYY-MM-DD HH:MM:SS` form must not be run through `new Date()`, because a string with no timezone is parsed as local time — west of UTC that moves the cursor forward, past rows that were never synced. Silent data loss rather than the redundant re-fetch you get going the other way. Your normalisation may well already sidestep this; flagging it because it cost me a test to notice.

No rush from my side, and I am not going to touch this one — it is yours. If you would rather I open the harness fix as a PR against your branch, say the word.

kacperpietrzyk:
Rebased onto `dev-2.6.1` at 2bb5a27 and adapted to the new harness — both pushed. And consider the #1140 thing settled: the revert plus merging #1139 covers it, no need to circle back.

**The `new Date()` trap — checked, this doesn't hit it.** `normalizeSyncTimestamp()` is pure string work, `value.replace("T", " ").slice(0, 19)`, with no `Date` constructor anywhere on the path, so a value already in `YYYY-MM-DD HH:MM:SS` form comes back byte-identical instead of being reinterpreted as local time. Worth having raised, though — the premise it rests on turns out to be SQLite-specific, which is the last section below.

**Harness fix.** Dropping the `CREATE TABLE` blocks was most of it, but seeding into the migrated tables surfaced two things the hand-written definitions had been hiding:

- the harness runs `PRAGMA foreign_keys = ON`, and both `sync_tombstones.user_id` and `ssh_credentials.user_id` reference `users`, so the owning row has to be seeded first;
- the real `ssh_credentials.auth_type` is `NOT NULL` with no default, where the local copy had `DEFAULT 'password'`.

`exec` is awaited as well, since it only returns synchronously on SQLite. The assertions are unchanged. Kept as a separate commit so the delta against the version you already read is easy to isolate — happy to squash if you'd rather it land as one.

**Verification.** The run on this PR is sitting at `action_required`, since fork PRs need your approval to start, so I can't turn it green from here. Locally on the rebased branch:

- `npm run lint` 0 errors, `prettier --check` clean, `tsc --noEmit` clean, `npm run build` fine
- full suite: 191/191 files, 1343 passed, 1 skipped, 0 failed
- the `database-dialects` job against real Postgres 16 and MySQL 8 in Docker — `verify:dialect` passed on both, repository suite 256/256 on each, these four tests included. The `substr(replace(...))` form was chosen for portability, so it seemed worth confirming against real servers rather than asserting it.

**One thing out of scope, but next door to what you're building.** On Postgres and MySQL, `CURRENT_TIMESTAMP` follows the session timezone; on SQLite it is always UTC. `updated_at` is `text` on all three, and both generated schemas default it to `CURRENT_TIMESTAMP`, so the session-local wall clock is what lands in the column. Two default-driven inserts of the same instant on the migrated schema, on the same containers as above:

```
postgres, SET TIME ZONE 'America/Los_Angeles' ->  2026-07-31 11:00:48.527775-07
postgres, SET TIME ZONE 'UTC'                 ->  2026-07-31 18:00:48.527775+00

mysql,    SET time_zone = '-08:00'            ->  2026-07-31 10:00:55
mysql,    SET time_zone = '+00:00'            ->  2026-07-31 18:00:55
```

Normalising to 19 characters drops the offset along with the fraction, so the first row of each pair reads seven or eight hours older than it is and sorts below an ISO (UTC) cursor that should have picked it up — the same silent skip you described, arriving through the session timezone rather than through `new Date()`. MySQL has no offset in the stored text at all, so the two are indistinguishable after the fact. CI can't see any of this because both service containers run UTC.

Not a regression from this PR: before it, every `CURRENT_TIMESTAMP` row failed the comparison on every engine anyway. I'd rather not widen this diff for it, but it's cheap to defuse at write time (`timezone('UTC', now())` / `UTC_TIMESTAMP()`, or standardising on the ISO form) if you want it handled while you're in here.

**On the offer** — I'll keep this one, the diff is small enough. No need to spend a PR on it.

LukeGus:
LGTM after review, thanks.

PR #1139, opened 2026-07-29
fix: make sync deletions reach the other side

kacperpietrzyk:
## Summary

- register `POST /sync/tombstones` before the `POST /:entityType` wildcard that was swallowing it
- check whether the receiving side still holds a row using its full set rather than the incremental window
- pay for that check only on a pass that actually carries a deletion
- regression tests for the route order and for both branches of the guard

Deletions have never propagated. Two independent defects, both hidden until now behind
Termix-SSH/Support#1050 (with the cursor broken, no tombstone was ever pulled to begin with).

**The endpoint was unreachable.** Express matches in registration order, and `"tombstones"` is a
perfectly good value for `:entityType`, so every deletion push landed in the wildcard handler and
came back as `400 "Unknown entity type"`. The handler below it has never run:

```
$ curl -X POST /sync/tombstones -d '{"entityType":"sshCredentials","syncId":"82629ab4-..."}'
{"error":"Unknown entity type"}   [400]
```

The GET pair is unaffected — `/:entityType/tombstones` and `/:entityType` differ in segment count.

**The guard consulted the incremental window.** `localBySyncId` / `remoteBySyncId` are built from
`pullSide(..., since)`. A row deleted on one side and untouched on the other is by definition
absent from that window — the shape every ordinary deletion takes once the two sides have
converged — so the tombstone was skipped, and skipped again on each later pass while sliding out of
its own window.

The guard can't just be dropped: `POST /sync/tombstones` records a tombstone on the receiving side,
so an unconditional push would hand the other side a fresh tombstone to push back on the next pass,
and the two would trade deletions forever. Asking the receiving side what it still holds keeps the
original intent and terminates — once the row is gone, nothing is pushed.

Cost: an ordinary pass is unchanged (the check is skipped when there are no tombstones); a pass
carrying a deletion adds one unfiltered list per affected entity type.

## Validation

- `npm test -- --run src/backend/tests/database/routes/sync.test.ts src/ui/tests/electron/remote-sync-tombstones.test.ts` (10 passed)
- both new tests confirmed to fail against `dev-2.6.1` without the corresponding change
- `npm test` — 181 files, 1248 passed, 1 skipped
- `npm run type-check`
- `npm run lint` (0 errors; the 44 existing warnings only)
- `npm run build`
- end to end, with #1050's patch applied on top so tombstones are pulled at all: server image built
  from the combined tree on an isolated volume, real `electron/remote-sync.cjs` against it.
  A credential deleted on the server disappears on the client on the next pass; a credential
  deleted on the client disappears on the server; three consecutive passes report
  `lastError: null` and the tombstone counts stay flat on both sides, so nothing ping-pongs.

## Ordering

This is behaviourally dependent on Termix-SSH/Support#1050 — without it the tombstone pull returns
nothing, so this change has nothing to act on. The two touch different regions of `sync.ts` and
should merge in either order; happy to rebase if they conflict.

Closes Termix-SSH/Support#1051

### DIFF
--- a/electron/remote-sync.cjs
+++ b/electron/remote-sync.cjs
@@ -398,6 +398,15 @@ class RemoteSyncEngine {
     return data.rows || [];
   }
 
+  /**
+   * Every syncId a side currently holds, ignoring the incremental window.
+   * Used only to decide whether a deletion still has something to delete.
+   */
+  async pullSyncIds(baseUrl, token, entityType) {
+    const rows = await this.pullSide(baseUrl, token, entityType, null);
+    return new Set(rows.filter((row) => row.syncId).map((row) => row.syncId));
+  }
+
   async pullTombstones(baseUrl, token, entityType, since) {
     const url = `${baseUrl}/sync/${entityType}/tombstones${since ? `?since=${encodeURIComponent(since)}` : ""}`;
     const data = await this.fetchJson(url, token);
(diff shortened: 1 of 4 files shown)


ZacharyZcR:
Sorry about this one — I filed #1140 covering the same two defects 4.5 hours after you opened this and #1138, without checking whether anyone was already on it, and then merged it. That was my mistake, not a judgement on your work. Support#1051 even ends with "Patch on the way." and I read straight past it.

#1140 is reverted (#1146). Your two PRs are the ones that should land.

I verified this branch locally, rebased onto current `dev-2.6.1`: `tsc --noEmit` clean, full suite green — 1291 passed, 1 skipped. The checks shown on this PR predate both the revert and #1147, so they are stale rather than meaningful.

Merging it now. Your tombstone guard is the better design of the two, incidentally: consulting the receiving side's full set rather than the incremental window is the correct fix, and only paying for that lookup on a pass that actually carries a deletion is a nicer touch than what I had — I made the endpoint idempotent server-side instead, which works but changes the endpoint's semantics to get there.

#1138 needs one small thing before it can follow — details over there.

PR #1141, opened 2026-07-29
Use the configured default credential when importing Proxmox guests

ZacharyZcR:
Fixes Termix-SSH/Support#1045.

The Proxmox integration offers five default auth types — `password`, `key`, `credential`, `opkssh`, `none` — and a separate default credential picker. Choosing **SSH Key** (or Password) imported every guest with auth type `none`, which is what the reporter hit.

## Why

`resolveProxmoxImportAuth` had three branches:

```ts
if (defaultAuthType === "credential" || (!defaultAuthType && credentialId)) { ... }
if (defaultAuthType && SECRETLESS_AUTH_TYPES.has(defaultAuthType))  return { authType: defaultAuthType };
if (defaultAuthType && !SECRET_BACKED_AUTH_TYPES.has(defaultAuthType)) return { authType: defaultAuthType };
return { authType: "none" };
```

`password` and `key` match none of them and fall to the final return. So the two types the dropdown documents were exactly the two that got dropped, while an auth type the resolver had never heard of was passed through untouched.

## What changed

`none` was the right answer for a discovered guest with nothing else configured — the guest carries no password and no private key. It is the wrong answer when a **default credential is set**, because that credential holds precisely the secret the guest is missing. That is already how an unset default behaves (`!defaultAuthType && credentialId` → credential auth); selecting `key` just stopped it from getting there.

So `password`, `key`, and `credential` are now treated alike: use the credential if one is configured, fall back to `none` if not. Everything else needs no secret and is imported as configured.

That collapses `SECRET_BACKED_AUTH_TYPES` and `SECRETLESS_AUTH_TYPES` into one set and removes a branch — the secretless list no longer has to be enumerated, since "not secret-backed" is the same statement.

| default auth type | credential | before | after |
|---|---|---|---|
| unset | set | credential | credential |
| unset | none | none | none |
| `credential` | set / none | credential / none | unchanged |
| **`password`** | **set** | **none** | **credential** |
| **`key`** | **set** | **none** | **credential** |
| `password` / `key` | none | none | none |
| `opkssh`, `none`, `tailscale`, `vault` | any | passed through | unchanged |
| unrecognized | any | passed through | unchanged |

Only the two bold rows change.

## Both copies

The backend (`routes/proxmox.ts`) and the frontend (`components/proxmox/proxmox-import-auth.ts`) each carry their own copy of this decision. Both are fixed, and both are now held to the same matrix in tests — including the secretless types and an unrecognized one — so the next drift between them fails a test instead of silently importing hosts that cannot authenticate.

`npx tsc --noEmit`, eslint, and prettier are clean; `src/ui/tests` + `src/backend/tests/database/routes` pass (501 tests).

### DIFF
--- a/src/backend/database/routes/proxmox.ts
+++ b/src/backend/database/routes/proxmox.ts
@@ -248,7 +248,13 @@ function mergeTags(
     .join(",");
 }
 
-function resolveProxmoxImportAuth(
+// Auth types that need a secret the discovered guest does not carry. A
+// configured default credential supplies exactly that, so these resolve to
+// credential auth when one is set and fall back to "none" when it is not.
+// Everything else needs no secret and is imported as configured.
+const SECRET_BACKED_AUTH_TYPES = new Set(["password", "key", "credential"]);
+
+export function resolveProxmoxImportAuth(
   defaultAuthType: string | undefined,
   credentialId: number | null | undefined,
 ): {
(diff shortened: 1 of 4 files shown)


ZacharyZcR:
Superseded by Termix-SSH/Termix#1144, which fixes the same defect as part of a broader set of Proxmox-behind-a-jump-host issues (Termix-SSH/Support#1054).

I compared the two resolvers cell by cell — the truth tables are identical, and #1144 additionally patches the discovery timeouts, the concurrency contention, the `NOT NULL constraint failed: ssh_data.username` abort on RDP guests, jump-chain and credential inheritance, and adds SSE progress. No reason to land a subset of it separately.

Details and the two gaps I flagged there: Termix-SSH/Termix#1144#issuecomment-5140641038

PR #1143, opened 2026-07-29
Resolve the dialect in the repository factory instead of assuming sqlite

ZacharyZcR:
Fixes Termix-SSH/Support#282.

I went in expecting wiring to be missing. It is not — `initializeDatabase()` already branches to `initializeRemoteDatabase()`, `connectRemoteDatabase()` and `runRemoteMigrations()` are in place, the migrations are generated, and CI runs the repository suite against real Postgres 16 and MySQL 8 containers. What was left is one line, and it is the one that matters.

## The line

```ts
export function createCurrentRepositoryContext(): DatabaseContext {
  return {
    dialect: "sqlite",   // <-- regardless of DATABASE_DIALECT
    drizzle: getDb(),
  };
}
```

`dialect` is not decoration. `returning.ts` reads it in five places to decide whether it can ask for RETURNING, and `upsert()` reads it to choose between `onConflictDoUpdate` and `onDuplicateKeyUpdate`. With it pinned to `"sqlite"`, a MySQL deployment takes the SQLite branch and calls `onConflictDoUpdate` on a mysql2 insert builder, which has no such method.

The comment already sitting above that function spelled out the consequence, for the case it assumed could not happen:

> calling it is a TypeError rather than a rejected query

So **MySQL never worked outside the test harness.** Postgres survived by coincidence — it supports RETURNING and shares the conflict syntax, so the wrong branch happens to be the right one.

## Why three layers of verification all missed it

- `tests/database/repositories/test-support.ts:79` builds `{ dialect: this.dialect, drizzle: db }` itself
- `scripts/verify-dialects.mjs:45` builds `{ dialect, drizzle: db }` itself
- the CI matrix runs both of the above against real engines

Every one of them constructs a `DatabaseContext` by hand and never calls the factory the application uses. Three passes, three engines, all green, and none of them touched the function under test. A fixture that agrees with your assumption cannot falsify it.

## The change

Resolve the dialect from the environment, and test the **factory** rather than a hand-built context:

- defaults to `sqlite` when nothing is configured
- reports `sqlite` / `postgres` / `mysql` as configured
- installs the persist write hook only for `sqlite`
- an unsupported value throws instead of silently falling back

Reverting the one-line fix fails two of these, which I checked rather than assumed.

I also added a short section to `docs/database-backends.md` under "What is verified, and how" — that section claimed thorough coverage and was exactly where the blind spot lived.

## Scope

Only the dialect resolution. I checked the rest of the surface rather than taking it on trust:

- every other `resolveDatabaseDialect` / `needsExplicitPersist` branch (`db/index.ts`, `sqlite-foreign-keys.ts`, both crypto migrations, `data-crypto.ts`) already resolves correctly
- `getCurrentSettingValue` correctly reads the primed cache on remote engines and only touches `getSqlite()` on the SQLite branch
- backup **import** genuinely refuses on non-SQLite with a readable message, and that guard is reachable from `database.ts:1277`
- backup **export** reads through repositories and writes a fresh SQLite file, so it is engine-independent

`npx tsc --noEmit`, eslint and prettier are clean; `src/backend/tests` passes (957 passed, 1 skipped).

### DIFF
--- a/docs/database-backends.md
+++ b/docs/database-backends.md
@@ -189,6 +189,19 @@ Tested against PostgreSQL 16 and MySQL 8. **MariaDB is not a substitute for
 MySQL when testing** — it accepts DDL that MySQL 8 rejects, which has hidden a
 real defect here more than once.
 
+### What neither of them covers
+
+Both harnesses build a `DatabaseContext` of their own, so neither runs
+`createCurrentRepositoryContext()` — the one the application actually uses.
+That gap hid a hardcoded `dialect: "sqlite"` in it: every engine reported
+itself as SQLite at runtime while all three test passes stayed green, which on
+MySQL meant `upsert` reached for `onConflictDoUpdate` and died with a
+TypeError on the first write.
+
+Anything the factory decides from the dialect needs its own test against the
+factory. Asserting it through a hand-built context proves nothing about what
+runs in production.
+
 ## Known limits
 
 - The desktop app always uses SQLite. It embeds its own backend and cannot ship
--- a/src/backend/database/repositories/factory.ts
+++ b/src/backend/database/repositories/factory.ts
@@ -47,9 +47,21 @@ import { UserRepository } from "./user-repository.js";
 import { VaultProfileRepository } from "./vault-profile-repository.js";
 import { VaultTokenRepository } from "./vault-token-repository.js";
 
+/**
+ * The context every repository runs against.
+ *
+ * The dialect has to be resolved, not assumed: it is what `returning.ts` reads
+ * to decide whether it can ask for RETURNING, and whether an upsert spells
+ * itself `onConflictDoUpdate` or `onDuplicateKeyUpdate`. Reporting "sqlite"
+ * while connected to MySQL makes the second of those a TypeError on the first
+ * write.
+ *
+ * Both cross-dialect harnesses build a DatabaseContext themselves, so neither
+ * exercises this function — see tests/database/repositories/factory-context.
+ */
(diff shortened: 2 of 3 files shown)


ZacharyZcR:
Flagging this as a **2.6.1 release blocker**, since `dev-2.6.1` now carries `Bump version from 2.6.0 to 2.6.1`.

2.6.1 ships Postgres and MySQL support along with `docs/database-backends.md`, which documents the configuration, gives a Docker compose example, and states "Tested against PostgreSQL 16 and MySQL 8". Without this one-line change, a user who follows that document and configures MySQL hits a TypeError on the first upsert — `createCurrentRepositoryContext()` reports `sqlite` regardless of `DATABASE_DIALECT`, so `upsert()` calls `onConflictDoUpdate` on a mysql2 insert builder, which does not have it.

Postgres is unaffected: it supports RETURNING and shares the conflict syntax, so the wrong branch happens to be the correct one there. MySQL is the only engine that breaks, and it breaks immediately.

Shipping a documented backend that fails on first write is worse than not shipping it, which is the only reason I am pushing on this rather than letting it sit in the queue.

Two options, either is fine by me:

1. **Merge this** — one line in `factory.ts` plus a test that fails if it is reverted (I checked that it does).
2. **Hold MySQL out of 2.6.1** — drop MySQL from the docs and have `resolveDatabaseDialect()` reject `mysql` with "not supported yet", ship Postgres only. I am happy to open that PR instead if it is the preferred call this close to a release.

What I would not do is release the docs as they stand with the resolver as it stands.

PR #1144, opened 2026-07-29
Make Proxmox guest discovery and import reliable over a jump host

XtraLarge:
Hi team — first, thank you for Termix. It's a genuine pleasure to run, and digging through this corner of the code was a treat. 🙏

This PR grew out of trying to import Proxmox guests from a node that's only reachable through a **jump host**, with the guests living behind that same jump. Discovery connected fine and then a small chain of things got in the way. Each is minor on its own; together they're the difference between "the Proxmox import doesn't work behind a jump" and "it just works". The write-up below breaks it down problem by problem. (There's also a one-line commit that closes an unterminated `version` string in `package.json` on this branch, so it builds.)

Everything here has been running in production against two real Proxmox 9.2.5 clusters — one connected directly, one two hops behind a jump over the internet — and is covered by the existing suite plus a new unit test.

# Overview

- [x] Fixed: Proxmox discovery intermittently timing out over a jump host
- [x] Fixed: discovery returning no IPs (so nothing imports) on small/remote nodes
- [x] Fixed: RDP guests aborting the whole sync (`NOT NULL` on `username`)
- [x] Added: guests without a resolvable IP now import with a `0.0.0.0` placeholder
- [x] Added: jump-host chain + credential inherited from the source host on import
- [x] Added: an SSE discovery stream with live `n/N` progress + heartbeat

# The problems, and what each fix does

### 1. Discovery times out intermittently over a jump — a bit of a lottery

`execCommand` caps every `pvesh` call at 8s:

```ts
function execCommand(client, command, timeoutMs = 8000) { ... }
```

Over a jump chain a single `pvesh get /cluster/resources` measured **8.3s** on our cluster — right on the edge. Some runs squeak under, most don't, and the whole discovery fails with `Command timed out after 8000ms`.

**Fix:** raise the per-call budget — 25s for the core calls, 12s for the best-effort agent/interface lookups. The jump path now completes reliably instead of coin-flipping.

### 2. It connects, but no guest gets an IP, so nothing imports

With discovery succeeding, the import still reported "no IP found" for every guest — even though the configs carry static IPs (`net0: ...,ip=10.0.0.5/24`). The cause was concurrency: `resolveIp` fans out `CONCURRENCY = 6` `pvesh` calls, and on a small node each `pvesh` (a Perl process + API round-trip) contends hard. Measured on the node:

| concurrent `pvesh` | time per call |
|---|---|
| 1 | ~6s |
| 2 | ~10s |
| 3 | ~26s ⛔ (over the 25s budget) |
| 5 | ~22s |

At 6 concurrent, every call blew the budget → every IP came back empty → the sync skipped all of them.

**Fix:** `CONCURRENCY = 2`. Gentle on small/remote nodes, each call stays comfortably under budget. It's a little slower on large local clusters, but the new progress UI (below) makes that visible rather than mysterious.

### 3. RDP guests abort the entire sync

A guest matched as RDP is created with `username: null`, but `ssh_data.username` is `NOT NULL`, so the whole sync throws — on every scheduled run:

```
SqliteError: NOT NULL constraint failed: ssh_data.username
```

**Fix:** use `""` for RDP guests, mirroring what the normal host-create path already does via `asString`.

### 4. Guests without a resolvable IP disappear silently

A QEMU guest with no running guest agent has no discoverable IP, and was skipped outright — so it couldn't be created at all, and the user had no way to add it.

**Fix:** import it with a `0.0.0.0` placeholder they can edit. Re-sync preserves the manual value (`guest.ip || existing.ip`) and never clobbers it back to the placeholder. (This also fixes a small table glitch — the IP cell is now always rendered, showing "no IP", so rows without an IP stay aligned.)

### 5. Imported guests don't inherit the jump host or credential from the source

The guests sit behind the same jump as their Proxmox host and authenticate with the same credential — but the manual import created them with no `jumpHosts` and `authType: "none"`, i.e. unreachable.

- **jumpHosts:** the discovery result now carries the source host's chain, and the import attaches it to each guest. (Auto-sync already did this; this aligns the manual path.)
- **credential:** `resolveProxmoxImportAuth` only used the inherited `credentialId` when `defaultAuthType` was `"credential"`. Since the default is `"password"`, a credential-authenticated host produced `"none"`. It now uses an available credential regardless — while an explicit secretless choice (`none`/`opkssh`/`tailscale`/…) still wins.

### 6. No feedback during a long discovery

A discovery that legitimately takes 40s+ behind a jump looked hung, and the blocking `POST` fought the 30s axios/proxy timeouts.

**Fix:** an SSE endpoint `GET /proxmox/discover/stream`, modeled on the existing `/ssh/tunnel/status/stream`. It emits a 15s heartbeat plus per-guest progress, so the dialog shows `n/N` and the connection stays alive as long as it needs. `POST /discover` stays as a fallback.

# Tests

- **New** `src/ui/tests/proxmox/proxmox-import-auth.test.ts` — 6 cases pinning the credential-inheritance behaviour (the regression that produced `authType: "none"`), including that an explicit secretless choice still wins and that `null` credentials fall back to `none`.
- `type-check`, `eslint`, `prettier --check` and `build` are clean.
- Existing suite: **1189 tests pass** locally.
- Verified end-to-end on two real Proxmox 9.2.5 clusters (direct + jump): discovery completes, guests import with IP or `0.0.0.0` placeholder + jump chain + credential, RDP guests import, and the scheduled auto-sync no longer errors.

# Checklist

- [x] This is not a translation request

Thanks so much for taking a look — happy to split this into smaller PRs, adjust the concurrency default, or add more tests if you'd prefer. 🙌


ZacharyZcR:
I had an open PR for the auth-type half of this — Termix-SSH/Termix#1141, filed against Termix-SSH/Support#1045 before #1054 came in. I went through yours and reached the same diagnosis independently, so this is a confirmation plus two gaps, not a competing patch. **I am closing #1141 in favour of this one**; yours is a strict superset and fixes five other things besides.

I compared the two resolvers cell by cell and they are behaviourally identical:

| defaultAuthType | credentialId | result |
|---|---|---|
| unset | set / unset | credential / none |
| `credential` | set / unset | credential / none |
| `password`, `key` | set / unset | **credential** / none |
| `none`, `opkssh`, `tailscale`, `vault` | any | passed through |
| unrecognized | any | passed through |

Mine collapsed the two type sets into one and dropped a branch; yours keeps three and puts the secretless check first. Same truth table, and yours states the secretless-wins precedence more explicitly — I would keep yours.

Two things worth picking up:

**1. The backend copy is changed but untested.** You correctly patched `resolveProxmoxImportAuth` in `src/backend/database/routes/proxmox.ts` as well, which matters because scheduled auto-sync goes through that one, not the dialog. But the new test file only covers the UI copy. The two implementations drifting apart is precisely what produced this bug, so the backend one is worth locking down too — export it and assert the same matrix. That is what `src/backend/tests/database/routes/proxmox-import-auth.test.ts` in #1141 does; feel free to lift it verbatim, no attribution needed.

**2. There will be two test files for the same function.** `src/ui/tests/components/proxmox/proxmox-import-auth.test.ts` already exists on `dev-2.6.1` and is not in this PR's diff, so adding `src/ui/tests/proxmox/proxmox-import-auth.test.ts` leaves both. The old one still passes (it only asserts the no-credential cases), so nothing breaks — it is just two files testing one function, and the older one is now the less complete of the two. Probably worth deleting it or landing your cases into it.

Nice write-up on the other five, incidentally — the `NOT NULL constraint failed: ssh_data.username` on RDP guests aborting the whole scheduled sync is a good catch.

XtraLarge:
Thanks so much for taking the time to go through this so carefully — genuinely appreciated, and it was a real pleasure working through it with you. Closing #1141 in favour of this one was very generous, and lifting your matrix made the follow-up much smoother. 🙏

Both gaps are addressed now:

**1. Backend copy is tested.** Importing the resolver straight from `proxmox.ts` drags the whole backend graph into the test env, so I extracted the decision into a leaf module `src/backend/database/routes/proxmox-import-auth.ts` (mirroring the UI's `proxmox-import-auth.ts`) and added `src/backend/tests/database/routes/proxmox-import-auth.test.ts` asserting the shared matrix. Your matrix carried over almost verbatim — thank you for offering it.

**2. Duplicate test removed.** Folded the fuller cases into the existing `src/ui/tests/components/proxmox/proxmox-import-auth.test.ts` and dropped `src/ui/tests/proxmox/…`, so there's a single file at the path mirroring the component.

A small bonus your review led to directly: the two copies still diverged on one real auth type — `agent`. The UI's secretless allowlist was missing it, so `resolveProxmoxImportAuth("agent", <credentialId>)` returned `credential` in the UI but passed `agent` through in the backend. Since agent auth is secretless, the backend had it right; I added `agent` to the UI set and both tests now cover the agent-with-credential case. Honestly, that one would likely have slipped through without your push for a shared matrix.

Thanks again — really enjoyed collaborating on this. 🙌

LukeGus:
LGTM after review, thanks.

PR #1149, opened 2026-08-01
fix(hosts): parse portKnockSequence JSON in host-resolver

XtraLarge:
## Problem

`resolveHostById` in `src/backend/hosts/host-resolver.ts` JSON-parses every serialized column — `jumpHosts`, `tunnelConnections`, `statsConfig`, `terminalConfig`, `socks5ProxyChain`, `quickActions` — **except** `portKnockSequence`.

An empty port-knock config is persisted as the **string** `"[]"` (the UI saves an empty array). Because `portKnockSequence` is never parsed, it stays a string, and the terminal code then evaluates `resolvedHostData.portKnockSequence.length > 0` on that string:

- `"[]".length === 2` → truthy → the backend logs `Loaded 2 port knock(s) from server-side host data` and `Port knocking <ip> (2 ports)` and attempts a knock for a host that has **no** knock configured.
- The `(2 ports)` is not a knock count — it's the character length of `"[]"`.

For a **real** knock sequence the same omission bites harder: the JSON string is never turned into the `Array<{ port, protocol?, delay? }>` that `performPortKnocking` iterates, so a genuine knock sequence would iterate string characters (`Number(char)` → `NaN` → skipped) and **never actually knock** — the host would silently fail to connect on setups that require knocking.

## Fix

Parse `portKnockSequence` like the sibling JSON columns:

- `"[]"` → `[]` → `length === 0` → no bogus log, no knock.
- a real sequence string → the array `performPortKnocking` expects.

## Tests

Adds two unit tests to `host-resolver.test.ts` covering both cases (`"[]"` → `[]`, and a real JSON sequence → parsed array). Full suite green (13/13); `tsc -p tsconfig.node.json` clean.

### DIFF
--- a/src/backend/hosts/host-resolver.ts
+++ b/src/backend/hosts/host-resolver.ts
@@ -113,6 +113,13 @@ export async function resolveHostById(
       host.quickActions = [];
     }
   }
+  if (typeof host.portKnockSequence === "string" && host.portKnockSequence) {
+    try {
+      host.portKnockSequence = JSON.parse(host.portKnockSequence as string);
+    } catch {
+      host.portKnockSequence = [];
+    }
+  }
 
   if (!ownerEquivalent) {
     const resolved = await resolveSharedSshSecrets(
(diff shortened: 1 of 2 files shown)


LukeGus:
LGTM after review, thanks.

PR #1157, opened 2026-08-04
fix tmux UTF-8 and non-login PATH handling

CarlHavingFun:
## Summary

- Preserve the inherited `PATH` as one quoted value in the non-login shell wrapper.
- Start every tmux client with `-u` so UTF-8 mode does not depend on the remote locale.
- Add focused regression coverage for path prefixes, shell escaping, command wrapping, and tmux detection.

## Root cause

The remote command wrapper expanded the inherited `PATH` without quotes, which could break command discovery when a path component contained whitespace. It also started tmux without forcing UTF-8 mode, so locale-limited remote sessions could create a non-UTF-8 client and corrupt non-ASCII terminal output.

## Impact

Tmux discovery continues to cover common non-login installation paths while preserving the complete inherited path. Attached and newly created tmux sessions consistently use UTF-8 mode.

## Validation

- `npx vitest run src/backend/tests/hosts/tmux/helper.test.ts` (4 tests)
- Prettier check on the changed files
- ESLint on the changed files
- `npm run type-check`
- `git diff --check`

### DIFF
--- a/src/backend/hosts/tmux/helper.ts
+++ b/src/backend/hosts/tmux/helper.ts
@@ -22,12 +22,12 @@ const TMUX_PATH_DIRS = [
 ];
 
 export function withTmuxPath(command: string): string {
-  const script = `PATH=${TMUX_PATH_DIRS.join(":")}:$PATH; export PATH; ${command}`;
+  const script = `PATH=${TMUX_PATH_DIRS.join(":")}:"$PATH"; export PATH; ${command}`;
   return `/bin/sh -c ${shellEscape(script)}`;
 }
 
 export function tmuxCommand(args: string): string {
-  return withTmuxPath(`tmux ${args}`);
+  return withTmuxPath(`tmux -u ${args}`);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
Leaving this as a comment rather than a review since it is still a draft — flagging one thing now so it does not bite you later, and approving CI so you can see the suite run.

The change itself looks right on both counts: quoting `"$PATH"` stops a path with spaces being word-split, and `tmux -u` is the correct fix for the mojibake. Tightening the assertions from `toMatch`/`toContain` to exact `toBe` strings is a real improvement — the old regexes would have passed even with `-u` missing.

One portability snag in the new test:

```js
expect(execFileSync("/bin/sh", ["-c", command], { encoding: "utf8" })).toBe("can't");
```

This shells out to a hardcoded `/bin/sh`. CI runs on ubuntu so it will pass there, but `npm test` on a Windows dev machine will fail on this one case — and Windows is a supported platform for the desktop app, so contributors do run the suite there.

Asserting on the escaped string itself keeps the same coverage without needing a shell:

```js
expect(withTmuxPath(`printf '%s' "can't"`)).toBe(
  `/bin/sh -c 'PATH=...:"$PATH"; export PATH; printf '"'"'%s'"'"' "can'"'"'\\'"'"''"'"'t"'`,
);
```

or gate it with `it.skipIf(process.platform === "win32")` if you would rather keep the round-trip through a real shell.

No rush — mark it ready when you are and I will review properly.

PR #1158, opened 2026-08-04
Send alerts in Discord channels with Webhooks

Horziox:
# Overview
Adds Discord webhook-based notification channels to Termix so alerts can be posted to Discord channels using a webhook URL. Keeps existing alert/channel behavior intact and improves outbound request safety and logging.

- [x] Added: Discord webhook sender and UI support for Discord channels
- [x] Updated: alert dispatching, outbound fetch safety, logger context
- [ ] Removed: n/a
- [x] Fixed: issues with undici/Agent-based outbound fetch (Node 24 compatibility) and client TypeScript typing for channel config

# Changes Made

_Detailed explanation of changes (if needed)_

- Backend

  - Added a dedicated Discord webhook sender with embed formatting, retry behavior and richer error logging: src/backend/utils/discord-sender.ts
  - Reworked outbound HTTP safety logic to avoid undici/Agent lookup quirks — DNS lookup + private address blocking, then use global fetch: src/backend/utils/safe-outbound-fetch.ts
  - Improved logger context formatting to include sanitized fields (url, error, truncated response bodies): src/backend/utils/logger.ts
  - Accept and validate "discord" channel type in notification channel routes and test endpoint: src/backend/database/routes/alert-rules-routes.ts
  - Route discord channels through the alert engine so alerts are dispatched to Discord webhooks: `src/backend/hosts/metrics/alert-engine.ts`
- Frontend / UI

  - Add Discord option to the Notification Channel dialog and fields for webhook URL, optional username and avatar URL: `src/ui/sidebar/NotificationChannelDialog.tsx`
  - Make the alerts API client accept structured config payloads (config: string | unknown) to match the UI behavior: `src/ui/api/alerts-api.ts`
- Tests / Validation

  - TypeScript compile performed and passes.
  - Unit tests for the safe-outbound-fetch helpers ran (vitest) and passed.
  - Manual runtime checks:
   - Backend starts successfully in Node 24 after native binding compatibility fixes.
   - Test delivery to Discord reaches Discord (invalid webhook returned 405) showing transport works and errors are actionable.

Notes on design decisions / TODOs:

- Storage format: Termix stores channel config as JSON strings in DB. The UI currently sends objects (config as structured object). This PR accepts unknown/object on the API side and server-side routes handle validation/stringification before persisting. Recommendation: standardize on either client-side `JSON.stringify(config)` or server-side normalization — the current approach is tolerant and backward-compatible.
- Further improvements possible: richer Discord message content, rate-limit (429) handling/backoff, and an integration test using a (mocked) webhook.

# Related Issues

_Link any issues this PR addresses_
n/a
# Screenshots / Demos

_(Optional: add before/after screenshots, GIFs, or console output)_

image
image

# Checklist


ZacharyZcR (changes requested):
Discord as an alert channel is a genuinely wanted feature and the sender itself reads well — the embed shape, the severity colour map, the retry. Two of the supporting changes have to come out before this can land, though, because both weaken existing protections rather than extend them.

## 1. `safe-outbound-fetch.ts` — the SSRF guard becomes checkable but not enforced

The current implementation resolves DNS **inside the connection** via an undici `Agent` with a `lookup` hook, so the addresses that get validated are the same ones the socket connects to. This PR replaces it with a standalone `lookup()` check followed by a plain `fetch()`:

```js
await new Promise(... lookup(hostname, ...) ...)   // resolves once, validates
return await fetch(url, { ...options, redirect: "error" })   // resolves again, connects
```

Those are two independent resolutions of the same name, which is exactly the window DNS rebinding targets:

```
lookup(hostname)  →  203.0.113.7   (public, passes)
fetch(url)        →  127.0.0.1     (attacker flips the record, connects anyway)
```

A webhook URL is user-supplied, so a hostile channel config is the expected threat here, not a hypothetical one. `createDnsLookupHook()` exists precisely to close that gap, and after this change it has no remaining caller in `src/backend` — only the test file references it.

Worth flagging that CI will not catch this. `safe-outbound-fetch.test.ts` exercises `createDnsLookupHook` directly against a fake resolver and deliberately avoids going through `Agent`/`fetch` (its own comment explains why), so the suite stays green while the guard is no longer wired into the request path.

If the `dispatcher` option was causing a type or runtime problem, could you say what you hit? That is worth solving on its own terms — the fix should keep validation and connection on one resolution.

## 2. `logger.ts` — turns log output from a whitelist into a blacklist

The formatter currently prints seven named fields. This change prints **every** remaining key in the context, minus the names in `SENSITIVE_FIELDS`. That inverts the default: a field is now exposed unless someone remembered to list it.

`SENSITIVE_FIELDS` (logger.ts:45-63) matches on exact key names, and `url` is not among them — while `discord-sender.ts` logs exactly that on failure:

```js
statsLogger.warn("Discord notification delivery failed after retry", {
  operation: "discord_notification_send_failed",
  url,                       // https://discord.com/api/webhooks/<id>/<token>
  ...
});
```

A Discord webhook URL *is* the credential — anyone holding it can post to the channel. So the first delivery failure writes it to the log in cleartext. Because the matching is by exact name, `webhookUrl`, `apiKey` and `refreshToken` would sail through too, at every call site in the codebase, not just this one.

The sender does not need the logger change at all: log the host, or add `url` to `SENSITIVE_FIELDS`. Please drop the `logger.ts` edit from this PR — if richer context output is worth having, it deserves its own PR where the whitelist/blacklist tradeoff can be argued on its own.

## 3. Formatting

The new block in `logger.ts` is indented with five spaces, so `npx prettier --check .` fails and takes `lint-and-build` with it. `npx prettier --write src/backend/utils/logger.ts` sorts it out — note the repo pins prettier 3.9.6, so make sure that is the version you run.

---

None of this is far from mergeable: revert the `safe-outbound-fetch.ts` and `logger.ts` changes, keep `url` out of the log line, and the feature itself stands on its own. I have approved CI so you can see the rest of the suite. Happy to re-review as soon as you push.

Horziox:
> Discord as an alert channel is a genuinely wanted feature and the sender itself reads well — the embed shape, the severity colour map, the retry. Two of the supporting changes have to come out before this can land, though, because both weaken existing protections rather than extend them.
> 
> ## 1. `safe-outbound-fetch.ts` — the SSRF guard becomes checkable but not enforced
> The current implementation resolves DNS **inside the connection** via an undici `Agent` with a `lookup` hook, so the addresses that get validated are the same ones the socket connects to. This PR replaces it with a standalone `lookup()` check followed by a plain `fetch()`:
> 
> ```js
> await new Promise(... lookup(hostname, ...) ...)   // resolves once, validates
> return await fetch(url, { ...options, redirect: "error" })   // resolves again, connects
> ```
> 
> Those are two independent resolutions of the same name, which is exactly the window DNS rebinding targets:
> 
> ```
> lookup(hostname)  →  203.0.113.7   (public, passes)
> fetch(url)        →  127.0.0.1     (attacker flips the record, connects anyway)
> ```
> 
> A webhook URL is user-supplied, so a hostile channel config is the expected threat here, not a hypothetical one. `createDnsLookupHook()` exists precisely to close that gap, and after this change it has no remaining caller in `src/backend` — only the test file references it.
> 
> Worth flagging that CI will not catch this. `safe-outbound-fetch.test.ts` exercises `createDnsLookupHook` directly against a fake resolver and deliberately avoids going through `Agent`/`fetch` (its own comment explains why), so the suite stays green while the guard is no longer wired into the request path.
> 
> If the `dispatcher` option was causing a type or runtime problem, could you say what you hit? That is worth solving on its own terms — the fix should keep validation and connection on one resolution.
> 
> ## 2. `logger.ts` — turns log output from a whitelist into a blacklist
> The formatter currently prints seven named fields. This change prints **every** remaining key in the context, minus the names in `SENSITIVE_FIELDS`. That inverts the default: a field is now exposed unless someone remembered to list it.
> 
> `SENSITIVE_FIELDS` (logger.ts:45-63) matches on exact key names, and `url` is not among them — while `discord-sender.ts` logs exactly that on failure:
> 
> ```js
> statsLogger.warn("Discord notification delivery failed after retry", {
>   operation: "discord_notification_send_failed",
>   url,                       // https://discord.com/api/webhooks/<id>/<token>
>   ...
> });
> ```
> 
> A Discord webhook URL _is_ the credential — anyone holding it can post to the channel. So the first delivery failure writes it to the log in cleartext. Because the matching is by exact name, `webhookUrl`, `apiKey` and `refreshToken` would sail through too, at every call site in the codebase, not just this one.
> 
> The sender does not need the logger change at all: log the host, or add `url` to `SENSITIVE_FIELDS`. Please drop the `logger.ts` edit from this PR — if richer context output is worth having, it deserves its own PR where the whitelist/blacklist tradeoff can be argued on its own.
> 
> ## 3. Formatting
> The new block in `logger.ts` is indented with five spaces, so `npx prettier --check .` fails and takes `lint-and-build` with it. `npx prettier --write src/backend/utils/logger.ts` sorts it out — note the repo pins prettier 3.9.6, so make sure that is the version you run.
> 
> None of this is far from mergeable: revert the `safe-outbound-fetch.ts` and `logger.ts` changes, keep `url` out of the log line, and the feature itself stands on its own. I have approved CI so you can see the rest of the suite. Happy to re-review as soon as you push.

Thanks a lot for the detailed review and for taking the time to point all of this out — the explanations were really helpful.

I’ve addressed all of the points you raised:

* Reworked `safe-outbound-fetch.ts` so the SSRF protection is enforced again on the actual connection path. I kept the DNS lookup hook wired into `undici.fetch` through the `Agent`/`dispatcher`, rather than doing a standalone DNS lookup followed by a separate `fetch`. This ensures that the IP addresses being validated are the same ones used for the connection, avoiding the DNS rebinding window you pointed out.
* I had initially tried to move away from the `undici` dispatcher, but that introduced DNS resolution issues (notably Discord endpoints were no longer resolving correctly). I therefore kept `undici.fetch` and fixed the underlying DNS lookup typing/handling instead, including the handling of public IPv4/IPv6 addresses, private/reserved addresses, IPv4-mapped IPv6 addresses, and invalid/empty lookup results.
* Removed the `logger.ts` changes entirely, as suggested. The Discord webhook URL is no longer exposed in the logs.
* Fixed the formatting issue as well.

I also added/fixed the relevant tests around the DNS lookup and SSRF checks.

TypeScript builds are passing, and the full test suite is passing as well.

Thanks again for the review and for the detailed context — it made it much easier to identify what needed to be changed without weakening the existing security guarantees.

ZacharyZcR (changes requested):
This is a substantial improvement — thank you for taking it seriously, and for answering the dispatcher question rather than just working around it.

Confirmed fixed:

- **SSRF enforcement is back on the connection path.** `createDnsLookupHook` is wired into the `Agent` again, so validation and connection share one resolution. The `all: false` branch you added is very likely what was breaking Discord before: the old hook forced `all: true` into the resolver and then handed back an array where undici expected a single address. Handling both shapes, plus stripping IPv6 brackets and rejecting a non-IP result, is a real fix rather than a workaround.
- **The logger change is gone**, and `"url"` in `SENSITIVE_FIELDS` is the right way to get the same protection. The webhook token no longer reaches the log.

Two things still need attention, one of them a straightforward bug.

## 1. The dispatcher is never closed

```diff
-  try {
-    return await fetch(url, { ...options, redirect: "error", dispatcher });
-  } finally {
-    await dispatcher.close();
-  }
+  return await undiciFetch(url.toString(), { ...options, dispatcher, redirect: "error" });
```

Every call constructs `new Agent(...)` and now nothing disposes of it, so each one leaks its connection pool and sockets.

This matters more than it might look, because `safeOutboundFetch` is not only the Discord path — `audit-forwarder.ts` calls it for **every forwarded audit entry**, and `notification-sender.ts` for every alert. On a busy instance that is a steady leak.

The `try/finally` needs to come back. Note it must not close a pooled proxy agent though, so something like: keep the agent in a local, and only close it in `finally` when it is the one this call created.

## 2. Routing through a proxy silently disables the guard

```js
const proxy = getProxyAgent(url.toString());
const dispatcher = proxy ?? new Agent({ connect: { lookup: createDnsLookupHook(lookup) } });
```

`getProxyAgent` returns a bare `new ProxyAgent(proxyUrl)` with no lookup hook, so when a proxy is configured the DNS validation does not run at all. The static check above it only catches literal IPs:

```js
if (isIP(hostname) && isBlockedAddress(hostname)) // a hostname skips this entirely
```

So with `HTTPS_PROXY` set, a webhook pointing at an internal name reaches the proxy unchecked — and a proxy is exactly what a corporate deployment has.

Proxy support may well be worth having, but it is a separate concern from adding a Discord channel, and it needs its own answer for how the destination gets validated when the client is not the one resolving it. Could you drop the `getProxyAgent` call from this PR? It restores the previous behaviour, and proxy support can land on its own with that question addressed properly.

## Minor

`"url"` in `SENSITIVE_FIELDS` is global and matches by exact key name, so every other log line carrying a `url` context field now prints `[MASKED]`. That is the correct trade here, but worth being aware of if some debug output gets less useful.

---

Fix those two and I will approve. The DNS handling itself I am happy with — it is more careful than what it replaced.

Horziox:
Thanks for the thorough review and for pointing these out !

Both issues are now fixed:

* The dispatcher is kept local and closed in a `finally` block after each request, so the per-call `Agent` does not leak its connection pool/sockets.
I initially assumed that once the Agent became unreachable, Node's garbage collector would eventually take care of its resources. That's why I removed the finally and went with a direct return :/
* I removed the `getProxyAgent` path from this PR, so the DNS/SSRF validation remains enforced on the connection path. Proxy support can be handled separately with the validation behaviour properly addressed.

I also addressed the minor point: I removed `"url"` from `SENSITIVE_FIELDS` since making it global was unnecessarily masking unrelated URL fields in logs. The webhook token is still protected through the appropriate logging/redaction path.

The tests have been updated and are passing.

PR #1159, opened 2026-08-04
Fix data export failing on Docker deployments

xDaly:
# Overview

Fixes user data export (SQLite) failing on Docker/localhost deployments — requests either crashed the backend or never reached it, depending on how the app was accessed.

- [ ] Added: ...
- [ ] Updated: ...
- [ ] Removed: ...
- [x] Fixed: User data export throwing `TypeError: Cannot read properties of undefined (reading 'username')`, and export/import requests being sent to an unreachable port on Docker deployments

# Changes Made

- `src/backend/database/database.ts`: `userRepository.findById()` returns a single `UserRecord | null`, not an array, but the `/database/export` route indexed it as `user[0]` in two places (building the export filename and building the exported `users` row). This was always `undefined`, causing the export request to crash with a 500 before any file was produced. Changed both to use `user` directly.
- `src/ui/lib/database-transfer-url.ts`: the dev-mode check treated *any* request from `hostname === "localhost"` or `"127.0.0.1"` as local development and forced the request straight to `http://localhost:30001/...`, bypassing nginx. This works only when port 30001 is published directly to the host (e.g. the raw `dev:docker` npm script), but breaks Docker Compose deployments (e.g. `docker/compose-dev.yml`) that only publish the nginx port, causing `net::ERR_CONNECTION_REFUSED`. Narrowed the check to the actual Vite dev server port (`5173`) only, so all other deployments correctly route through nginx like the rest of the app's API calls already do.

# Related Issues

- Closes #ISSUE_NUMBER

# Screenshots / Demos

image

image

# Checklist

### DIFF
--- a/src/backend/database/database.ts
+++ b/src/backend/database/database.ts
@@ -713,7 +713,7 @@ app.post("/database/export", authenticateJWT, async (req, res) => {
     }
 
     const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
-    const filename = `termix-export-${user[0].username}-${timestamp}.sqlite`;
+    const filename = `termix-export-${user.username}-${timestamp}.sqlite`;
     const tempPath = path.join(tempDir, filename);
 
     apiLogger.info("Creating export database", {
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
Thanks for tracking this down — both halves of it are already on `main`, so there is nothing left for this branch to carry.

They landed in `a64c956c` ("release-2.6.1", #1161):

- `src/backend/database/database.ts:716` is already `user.username` — the `user[0]` indexing went away with the repository-layer change that made the lookup return a record rather than a row array.
- `src/ui/lib/database-transfer-url.ts:23` is already `const development = location.port === "5173"`, so a Docker deployment reached over `localhost` no longer misroutes the export to `localhost:30001`.

Closing as superseded rather than because the diagnosis was wrong — it was right, and you got to it independently. If you are still seeing an export failure on 2.6.1 or later, please open a fresh issue with the server log line; that would be a different bug from this one.

PR #1167, opened 2026-08-07
fix(desktop): stop suppressing the update prompt, and make the version badge reachable

kacperpietrzyk:
# Overview

The desktop app's startup update modal is suppressed for exactly the users it exists for, and the "UPDATE AVAILABLE" badge that remains is not clickable. Together they make a new release unreachable from inside the app.

- [x] Fixed: update prompt keyed on the offered remote version instead of the local one
- [x] Fixed: version badge links to the release on both surfaces that render it
- [x] Added: `components/version-badge.tsx`, extracted from two duplicated copies
- [x] Added: tests for both (there were none)

# Changes Made

### The modal never reappears once you are behind

`src/ui/user/ElectronVersionCheck.tsx` stored its dismissal under the **local** app version:

```js
const dismissedVersion = localStorage.getItem("electron-version-check-dismissed");
if (dismissedVersion === currentVersion) { onContinue(); return; }
...
if (updateInfo?.status === "up_to_date") {
  localStorage.setItem("electron-version-check-dismissed", currentVersion);
```

`currentVersion` is `app.getVersion()`, and the `up_to_date` branch writes it with **no user interaction at all**. So anyone who launches the app while current has their own version recorded. When the next release ships, `dismissedVersion === currentVersion` still holds, the modal is skipped on every launch, and it only comes back *after* they have already updated — the inverse of its purpose. This has been the behaviour since v2.3.0.

Confirmed on a real install: the app's leveldb held `electron-version-check-dismissed = 2.6.0` for the whole period 2.6.1 was available, then `= 2.6.1` after a manual update.

This PR keys the dismissal on the remote version being offered. It is backward compatible — an existing key holding `2.6.0` compares unequal against a remote `2.6.1`, so affected installs get prompted on their next launch, with no migration step. When the check fails there is no remote version, so nothing is written and no future prompt is suppressed.

### The badge was the only thing left, and it was inert

With the modal silent, the version badge is the sole signal a release exists. It was a plain `<span>` — no href, no handler, no tooltip — even though the `getVersionInfo()` response behind it already carries `latest_release.html_url`.

There were two near-identical copies of that badge, in `sidebar/UserProfilePanel.tsx` and in `StatsBarCard` in `dashboard/DashboardTab.tsx`. They are now one `VersionBadge` component. In the `requires_update` case it renders an anchor to the release with `target="_blank"`, `rel="noopener noreferrer"`, and an accessible name that says where it leads, since "UPDATE AVAILABLE" alone does not. The `beta` and stable cases are unchanged inert spans, and appearance is preserved on both surfaces (the dashboard's `w-fit` comes through a `className` passthrough).

`getVersionInfo()` returned `Record<string, unknown>`, so the release URL was not reachable without a cast. It now returns a `VersionInfo` interface. The index signature on it is deliberate: `SystemOverviewWidget` reads `updateAvailable` off this response even though `GET /version` does not return such a field, so the type stays as permissive as the one it replaced rather than turning an unrelated stale read into a build error. Worth a separate look — that widget also calls `getVersionInfo(false)`, and `checkRemote=false` short-circuits before any remote comparison, so its update indicator cannot fire either way.

### Tests

Neither `ElectronVersionCheck` nor the badge had any tests. Added 19 across `tests/user/ElectronVersionCheck.test.tsx`, `tests/components/version-badge.test.tsx` and `tests/api/system-status-api.test.ts`, covering dismissal keying, the up-to-date and check-failed paths, the beta channel, the badge's link/inert cases, and the read that supplies the URL.

That last one is deliberate: the badge component is unit-tested, but the line deciding whether it ever gets a URL was duplicated at both call sites and asserted nowhere, and a wrong property there still compiles because the response type keeps an index signature. It is now `releaseUrlFrom`, used by both surfaces and tested directly.

The regression is genuinely captured — reverting only `ElectronVersionCheck.tsx` and re-running its suite fails 5 of 7:

```
 × still prompts when the stored dismissal is the local version and a newer release exists
 × stores the remote version when the user dismisses the prompt
 × skips the prompt on a later launch offering the same remote version
 × keys a dismissed beta prompt on the remote version too
 × continues without prompting when up to date, recording the offered version
 Tests  5 failed | 2 passed (7)
```

Verified against `main` @ `57b2081`:

| | before | after |
|---|---|---|
| `npm run type-check` | clean | clean |
| `npm run lint` | 0 errors, 42 warnings | 0 errors, 42 warnings |
| `npm test` | 202 files, 1453 passed / 1 skipped | 205 files, 1472 passed / 1 skipped |

### Scope

This does not make the app self-updating — the link still opens a browser. Real in-app updates need `electron-updater` plus a `publish` config and release-workflow changes, which is a separate and much larger change; it stays on Termix-SSH/Support#1087. Nothing under `electron/`, `electron-builder.json`, or `.github/workflows/` is touched here.

# Related Issues

- Related to Termix-SSH/Support#1087

# Screenshots / Demos

Not included — the change is a link affordance on an existing badge and a modal that was already implemented but unreachable. The test output above is the demonstrable part.

# Checklist

- [x] Supports mobile and desktop UI/app (if applicable) — both badge surfaces are desktop-shell views; no mobile view renders a version badge

One note on translations: the single new key is added to `src/ui/locales/en.json` only. Nothing under `src/ui/locales/translated/` is touched, so Crowdin stays authoritative.

### DIFF
--- a/src/ui/api/system-status-api.ts
+++ b/src/ui/api/system-status-api.ts
@@ -137,9 +137,10 @@ export interface VersionInfo {
   };
   cached?: boolean;
   cache_age?: number;
-  // The endpoint carries more than the fields above, and older callers read
-  // some of them (`updateAvailable`, for one), so stay a superset of the
-  // `Record<string, unknown>` this used to return.
+  // Callers reach for fields beyond the ones above -- SystemOverviewWidget
+  // reads `updateAvailable`, which this endpoint does not in fact return --
+  // so keep the index signature the previous `Record<string, unknown>` gave
+  // them. Typing those reads out of existence is a separate change.
   [key: string]: unknown;
 }
 
--- a/src/ui/api/system-status-api.ts
+++ b/src/ui/api/system-status-api.ts
@@ -124,9 +124,26 @@ export async function getReleasesRSS(
   }
 }
 
-export async function getVersionInfo(
-  checkRemote = true,
-): Promise<Record<string, unknown>> {
+export interface VersionInfo {
+  status?: "up_to_date" | "requires_update" | "beta";
+  localVersion?: string;
+  remoteVersion?: string;
+  latest_release?: {
+    tag_name?: string;
+    name?: string;
+    published_at?: string;
+    html_url?: string;
+    body?: string;
+  };
+  cached?: boolean;
+  cache_age?: number;
+  // The endpoint carries more than the fields above, and older callers read
+  // some of them (`updateAvailable`, for one), so stay a superset of the
+  // `Record<string, unknown>` this used to return.
+  [key: string]: unknown;
+}
+
+export async function getVersionInfo(checkRemote = true): Promise<VersionInfo> {
   try {
     const response = await authApi.get(
       `/version${checkRemote ? "" : "?checkRemote=false"}`,
--- a/src/ui/api/system-status-api.ts
+++ b/src/ui/api/system-status-api.ts
@@ -143,6 +143,14 @@ export interface VersionInfo {
   [key: string]: unknown;
 }
 
+// Where the release page lives inside a version response. Both surfaces that
+// render a version badge read it, and an empty string is what they treat as
+// "no link to offer", so keep the shape in one place rather than repeating the
+// optional chain at each call site.
+export function releaseUrlFrom(info: VersionInfo | null | undefined): string {
+  return info?.latest_release?.html_url ?? "";
+}
+
 export async function getVersionInfo(checkRemote = true): Promise<VersionInfo> {
   try {
     const response = await authApi.get(
--- a/src/ui/components/version-badge.tsx
+++ b/src/ui/components/version-badge.tsx
@@ -1,0 +1,52 @@
+import { useTranslation } from "react-i18next";
+
+interface VersionBadgeProps {
+  status: "up_to_date" | "requires_update" | "beta";
+  releaseUrl?: string;
+  className?: string;
+}
+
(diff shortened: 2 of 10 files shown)


ZacharyZcR:
I have reviewed this and the logic holds up — flagging one mechanical blocker so you can clear it.

`lint-and-build` fails on formatting only:

```
[warn] src/ui/tests/api/system-status-api.test.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

One file, nothing to do with the change itself:

```
npx prettier@3.9.6 --write src/ui/tests/api/system-status-api.test.ts
```

The version matters — the repo pins prettier 3.9.6 and 3.8.x disagrees with it about wrapping union types, so a globally installed prettier or an editor plugin bundling its own copy can leave the file failing `--check` while looking formatted locally. I lost a round to exactly that earlier today. (We are adding a pre-commit hook so this stops being something contributors have to know.)

On the substance: keying the dismissal on the remote version instead of the local one is the right fix, and the reasoning in the comment is correct — storing the local version meant the prompt was suppressed for precisely the users who had not updated yet, and reappeared only once it was no longer needed. Falling back to the local version when the check fails is a nice touch, since it can never match a later release. `VersionBadge` removes real duplication, `rel="noopener noreferrer"` and the spelled-out `aria-label` are right, and all four i18n keys resolve — three already existed and you added the missing `versionCheck.updateLinkLabel`.

Push the formatting fix and I will approve.

PR #1168, opened 2026-08-07
fix(homepage): make the System Overview update indicator able to fire

### LINKED ISSUE #1088: [BUG] System Overview widget's update indicator can never fire
### Title

System Overview widget's update indicator can never fire — it reads a field `/version` does not return, after asking `/version` not to check

### Platform

Web-app

Docker

### Version

2.6.1

### Troubleshooting

- [x] I have checked open issues and ensured this is not a duplicate

### The Problem

The **System Overview** homepage widget renders an "Update available" row and turns the version text orange when a newer release exists. That never happens, for two independent reasons that each on their own would be enough.

`src/ui/features/homepage/widgets/SystemOverviewWidget.tsx`:

```js
const info = await getVersionInfo(false);
setVersion((info?.localVersion as string) ?? null);
setUpdateAvailable(Boolean(info?.updateAvailable));
```

1. **`checkRemote=false` short-circuits before any comparison.** `src/backend/database/database.ts` returns early for that query parameter:

   ```js
   if (req.query.checkRemote === "false") {
     return res.json({ localVersion, status: "update_check_disabled" });
   }
   ```

   No GitHub fetch, no `remoteVersion`, no comparison. The widget asked the endpoint not to do the one thing it needs.

2. **`updateAvailable` is not a field this endpoint returns.** The success response is built as `{status, localVersion, version, remoteVersion, latest_release, cached, cache_age}` — there is no `updateAvailable` anywhere in the route. `Boolean(info?.updateAvailable)` is `Boolean(undefined)`, permanently `false`.

The read type-checks only because `getVersionInfo()` is declared as `Record<string, unknown>`, so any property name is accepted and nothing flags the typo-shaped access.

Verified against a live 2.6.1 server. `GET /version` returns:

```json
{"status":"requires_update","localVersion":"2.6.0","version":"2.6.1","remoteVersion":"2.6.1",
 "latest_release":{"tag_name":"release-2.6.1-tag","html_url":"https://github.com/Termix-SSH/Termix/releases/tag/release-2.6.1-tag"},
 "cached":true,"cache_age":114335}
```

`status` carries the answer. `updateAvailable` is absent, exactly as the route source says.

There is a third, cosmetic problem hiding behind the first two. The row is built as label `homepage.overviewUpdate` (English string: **"Up to date"**) with value `homepage.overviewUpdateAvailable` ("Update available"), so the moment the indicator does fire it reads:

```
Up to date        Update available
```

Nobody has seen that because the row has never rendered. Fixing the indicator without fixing the label would ship it.

### How to Reproduce

1. Run a server whose version is behind the latest GitHub release (e.g. 2.6.0 while 2.6.1 is out).
2. Add the **System Overview** widget to the homepage with "show version" enabled.
3. Observe: the version is rendered in the normal colour and no "Update available" row appears.
4. Compare with the dashboard stats bar or the user profile panel on the same server — both correctly show the "UPDATE AVAILABLE" badge, because they call `getVersionInfo()` without `checkRemote=false` and read `status`.

Confirm the cause directly:

```console
$ curl -s "$URL/version?checkRemote=false" -H "Authorization: Bearer $TOKEN"
{"localVersion":"2.6.1","status":"update_check_disabled"}
```

No field the widget could act on.

### Additional Context

The fix is small: drop `checkRemote=false` so the comparison actually runs, read `status === "requires_update"` like the other two surfaces do, and give the row a label that is not the string "Up to date". I have a PR ready and will link it here.

Two related notes for whoever reviews it:

- The other three version surfaces read `status`. This widget is the only one that invented a different field name, which is what makes it worth a type on the response rather than another one-off cast.
- Termix-SSH/Support#1087 and its PR cover a different set of version-surface problems (a startup update prompt that suppresses itself, and badges with no link). They do not overlap with this widget; this is a separate defect in a separate component.

kacperpietrzyk:
# Overview

The System Overview homepage widget has an update indicator that can never fire.

- [x] Fixed: widget asks `/version` to actually perform the remote check
- [x] Fixed: reads `status`, the field the endpoint returns, instead of one it does not
- [x] Fixed: the row is no longer labelled "Up to date"
- [x] Added: tests for the widget (there were none)

# Changes Made

`src/ui/features/homepage/widgets/SystemOverviewWidget.tsx` did this:

```js
const info = await getVersionInfo(false);
setVersion((info?.localVersion as string) ?? null);
setUpdateAvailable(Boolean(info?.updateAvailable));
```

Two independent faults, either of which alone would have been enough to make the indicator dead:

**`checkRemote=false` short-circuits the endpoint.** `src/backend/database/database.ts` returns before any remote work:

```js
if (req.query.checkRemote === "false") {
  return res.json({ localVersion, status: "update_check_disabled" });
}
```

**`updateAvailable` is not a field `/version` returns.** The success response is assembled as `{status, localVersion, version, remoteVersion, latest_release, cached, cache_age}`. `Boolean(undefined)` is `false`, permanently. Against a live 2.6.1 server:

```console
$ curl -s "$URL/version?checkRemote=false" -H "Authorization: Bearer $TOKEN"
{"localVersion":"2.6.1","status":"update_check_disabled"}

$ curl -s "$URL/version" -H "Authorization: Bearer $TOKEN" | jq 'keys'
["cached","latest_release","localVersion","remoteVersion","status","version"]
```

It type-checked only because `getVersionInfo()` is declared `Record<string, unknown>`, which accepts any property name — a field that does not exist is indistinguishable from one that does.

The fix lets the endpoint do its comparison and reads `status === "requires_update"`, which is what the dashboard stats bar and the profile panel badge already read off the same call.

### The label

The row was built as label `homepage.overviewUpdate` — English string **"Up to date"** — with value `homepage.overviewUpdateAvailable`. So the first thing it would have rendered once working is:

```
Up to date        Update available
```

No one has seen this, because the row has never rendered. Fixing the indicator without the label would have shipped it. The label now has its own key. `homepage.overviewUpdate` is deliberately left in place rather than deleted: "Up to date" is the natural *value* for an always-visible status row, and whether the widget should show one is a product decision rather than part of this fix.

Only `src/ui/locales/en.json` is touched; nothing under `src/ui/locales/translated/`.

### Tests

The widget had none. Added 5 covering both regressions and the label, plus the up-to-date and beta paths. Reverting only the component and re-running fails 3 of 5:

```
 × asks /version to actually check the remote release
 × flags an available update from the status the endpoint returns
 × labels that row 'Update' rather than 'Up to date'
 Tests  3 failed | 2 passed (5)
```

Verified against `main` @ `57b2081`:

| | before | after |
|---|---|---|
| `npm run type-check` | clean | clean |
| `npm run lint` | 0 errors, 42 warnings | 0 errors, 42 warnings |
| `npm test` | 202 files, 1453 passed / 1 skipped | 203 files, 1458 passed / 1 skipped |

### Relationship to #1167

Independent, and deliberately so — no shared files, so the two can merge in either order without conflict. #1167 fixes a different set of version surfaces (the startup update prompt suppressing itself, and the two badges having no link). This one is the third surface, broken a different way.

One sequencing note if both land: #1167 gives `getVersionInfo()` a `VersionInfo` return type whose permissive index signature exists specifically so *this* widget's `updateAvailable` read keeps compiling. Once this PR removes that read, nothing depends on the index signature any more and it can be tightened — worth a follow-up, not something either PR should do to the other.

# Related Issues

- Closes Termix-SSH/Support#1088

# Screenshots / Demos

Not included — the visible change is a row that previously could not render. The test output above is the demonstrable part.

# Checklist

- [x] Supports mobile and desktop UI/app (if applicable) — homepage widget, unchanged layout behaviour

### DIFF
--- a/src/ui/features/homepage/widgets/SystemOverviewWidget.tsx
+++ b/src/ui/features/homepage/widgets/SystemOverviewWidget.tsx
@@ -47,9 +47,14 @@ function SystemOverviewWidget({
   const fetchData = async () => {
     try {
       if (showVersion) {
-        const info = await getVersionInfo(false);
+        // `checkRemote=false` makes /version return early with just
+        // {localVersion, status: "update_check_disabled"}, and the response has
+        // no `updateAvailable` field in either mode -- so the previous read was
+        // false twice over. `status` is what the endpoint actually answers with,
+        // and what the dashboard and profile badges already read.
+        const info = await getVersionInfo();
         setVersion((info?.localVersion as string) ?? null);
-        setUpdateAvailable(Boolean(info?.updateAvailable));
+        setUpdateAvailable(info?.status === "requires_update");
       }
       if (showDbHealth) {
         const health = await getDatabaseHealth();
@@ -93,7 +98,7 @@ function SystemOverviewWidget({
         )}
         {showVersion && updateAvailable && (
           <InfoRow
-            label={t("homepage.overviewUpdate")}
+            label={t("homepage.overviewUpdateLabel")}
             value={t("homepage.overviewUpdateAvailable")}
             valueColor="#f97316"
           />
--- a/src/ui/locales/en.json
+++ b/src/ui/locales/en.json
@@ -211,6 +211,7 @@
     "uptimeLabel": "Uptime",
     "overviewVersion": "Version",
     "overviewUpdate": "Up to date",
+    "overviewUpdateLabel": "Update",
     "overviewUpdateAvailable": "Update available",
     "overviewDatabase": "Database",
     "overviewUptime": "Uptime",
--- a/src/ui/tests/features/homepage/SystemOverviewWidget.test.tsx
+++ b/src/ui/tests/features/homepage/SystemOverviewWidget.test.tsx
@@ -1,0 +1,126 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { render, screen, cleanup, waitFor } from "@testing-library/react";
+
+vi.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+const getVersionInfo = vi.fn();
+const getDatabaseHealth = vi.fn(async () => ({ status: "ok" }));
+vi.mock("@/api/system-status-api", () => ({
+  getVersionInfo: (...args: unknown[]) => getVersionInfo(...args),
+  getDatabaseHealth: () => getDatabaseHealth(),
+}));
+vi.mock("@/api/dashboard-api", () => ({
+  getUptime: vi.fn(async () => ({ formatted: "1d 2h" })),
+}));
+
+// The real one installs an interval and a visibilitychange listener; the
+// component's first fetch happens outside it, which is all these tests need.
+vi.mock("../../../features/homepage/use-visible-interval", () => ({
+  runVisibleInterval: () => () => {},
+}));
+
+import { SystemOverviewWidget } from "../../../features/homepage/widgets/SystemOverviewWidget";
+import type { CanvasWidget } from "@/types/homepage-types";
+
+const widget: CanvasWidget = {
+  id: 1,
+  typeId: "system_overview",
+  title: "Termix",
+  config: {},
+  x: 0,
+  y: 0,
+  w: 10,
+  h: 6,
+  zOrder: 0,
+};
+
+function renderWidget() {
+  render(
+    <SystemOverviewWidget
+      widget={widget}
+      config={{ showVersion: true, showDbHealth: false, showUptime: false }}
+    />,
+  );
+}
+
+beforeEach(() => {
+  getVersionInfo.mockReset();
+});
+
+afterEach(cleanup);
+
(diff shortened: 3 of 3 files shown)


ZacharyZcR (approved):
Verified all three fixes against the endpoint rather than taking them on trust:

- `getVersionInfo(false)` makes `/version` return at `database.ts:355` with `{localVersion, status: "update_check_disabled"}` — the remote was never consulted.
- The response has no `updateAvailable` field in either mode (`database.ts:378-395` returns `status`), so `Boolean(info?.updateAvailable)` was false regardless. Failing twice over, as your comment says.
- The label read `overviewUpdate`, whose value is "Up to date", so the row rendered as "Up to date: Update available".

Switching to `status === "requires_update"` is right, and it correctly leaves `beta` (local ahead of remote) without a prompt. The remote call goes through `fetchGitHubAPI` with a cache key, so moving to `checkRemote=true` on a polled widget does not put pressure on the GitHub API. Adding only `en.json` is correct here — the rest comes back through Crowdin.

Both checks are green. Thanks for the 126 lines of tests, and for explaining the double failure in a comment instead of just fixing it.

PR #1169, opened 2026-08-08
fix: capture real client IP for SSH login alerts behind reverse proxy

### LINKED ISSUE #1100: SSH login alert shows proxy/loopback IP instead of real client IP behind reverse proxy
## Description
When Termix runs behind a reverse proxy chain (e.g. Cloudflare Tunnel → Traefik → Termix), SSH login alerts still show a loopback/proxy address instead of the real client IP:

```
User "root" logged in to "Hyper" from ::ffff:127.0.0.1
```

This happened even after reconfiguring Traefik to trust forwarded proxies / set real client IP headers.

## Root cause
The WebSocket terminal handler (`src/backend/hosts/terminal/index.ts`) used `req.socket.remoteAddress` for the login-alert IP, which is the immediate TCP peer (the proxy), not the value from `X-Forwarded-For`. Reverse-proxy trust settings had no effect because Termix never read the forwarded header on this code path in the first place.

## Fix
A fix has been submitted: https://github.com/Termix-SSH/Termix/pull/1169

Adds a `getClientIp` helper (checks `X-Forwarded-For` → `req.ip` → raw socket, in that order) and wires it into the login-alert trigger and the WS auth-failure log.

## Environment
- Deployment: Cloudflare Tunnel → Traefik → Termix

brennanneoh:
## Summary
- The WebSocket terminal handler used `req.socket.remoteAddress` for the "user logged in" alert message, which is the immediate TCP peer (the reverse proxy) rather than the real client IP forwarded via `X-Forwarded-For`. Reconfiguring Traefik's trust-proxy settings had no effect because Termix never read the header on this code path.
- Added `getClientIp` in `request-origin.ts` (forwarded header → `req.ip` → raw socket, in that order) and wired it into both the login-alert trigger and the WS auth-failure log in `terminal/index.ts`.

Fixes Termix-SSH/Support#1100

## Test plan
- [x] `npx vitest run src/backend/tests/utils/request-origin.test.ts` — 17 passed
- [x] `npx tsc --noEmit`

### DIFF
--- a/src/backend/hosts/terminal/index.ts
+++ b/src/backend/hosts/terminal/index.ts
@@ -45,6 +45,7 @@ import {
 import { isWindowsSftpPath, sftpPathToLocalPath } from "../transfer-paths.js";
 import { preparePrivateKeyForSSH2 } from "../../utils/ssh-key-utils.js";
 import { triggerLoginAlert } from "../../utils/alert-trigger.js";
+import { getClientIp } from "../../utils/request-origin.js";
 import { isRetriableDnsError, resolveHostForSshConnect } from "../ssh-dns.js";
 
 interface ConnectToHostData {
@@ -326,7 +327,7 @@ wss.on("connection", async (ws: WebSocket, req) => {
       error,
       {
         operation: "websocket_connection_auth_error",
-        ip: req.socket.remoteAddress,
+        ip: getClientIp(req),
       },
     );
     ws.close(1008, "Authentication required");
@@ -2184,7 +2185,7 @@ wss.on("connection", async (ws: WebSocket, req) => {
               id,
               hostConfig.userId,
               username,
-              req.socket.remoteAddress ?? "unknown",
+              getClientIp(req),
             ).catch(() => {});
           }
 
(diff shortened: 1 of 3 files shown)


ZacharyZcR (approved):
Checked the pieces that mattered rather than assuming them:

- `firstHeaderValue` does `raw.split(",")[0].trim()`, so `"203.0.113.7, 10.0.0.1, 10.0.0.2"` yields the leftmost hop as your test expects — the array-vs-string cases both land correctly.
- The precedence order is right: forwarded header, then `req.ip`, then the raw socket peer, then `"unknown"`. Behind Traefik or Cloudflare the socket peer really is a loopback address, which is exactly the symptom in Support#1100.
- Trusting `X-Forwarded-For` matches the existing posture — `database.ts:66` already sets `app.set("trust proxy", true)` — so this is consistent rather than a new assumption.

Five cases including "no IP information at all" is thorough for a 14-line function. Both checks green.

One follow-up worth doing separately, not a blocker: `getRequestMeta` in `src/backend/utils/audit-logger.ts:59-66` has near-identical logic that is strictly worse — its array branch does not split or trim, it never falls back to the socket peer, and it returns an empty string rather than `"unknown"`. Now that `getClientIp` exists it should just call it, so the audit trail gets the same correctness. Happy to open that myself if you would rather not expand this PR.

Thanks for the clean separation into a reusable helper instead of inlining it at the two call sites.

PR #1170, opened 2026-08-08
feat: add terminal image handoff

T3rM1nAt0-R:
## Summary

Adds a browser image handoff flow for connected Termix terminals:

- Upload an image from the file picker.
- Paste an image from the browser clipboard.
- Normalize accepted image bytes through Sharp and store them as PNG.
- Insert a reviewable `Please inspect this image: <host path>` prompt into the
  active terminal input without submitting it automatically.
- Keep image storage bounded with TTL, file-count, byte-size, and decoded-pixel
  limits.
- Support explicit container/host path mapping through:
  - `TERMIX_IMAGE_DIR`
  - `TERMIX_IMAGE_HOST_PATH`
  - `TERMIX_IMAGE_TTL_MS`
- Add a 50 MB request limit to both HTTP and HTTPS nginx configurations.
- Keep the terminal corner controls in one stable three-button group:
  `tmux:detach`, `upload image`, and `paste image`.
- Add dev API proxy support for the backend service ports.

## Reliability details

- Browser MIME metadata is not trusted for validation; Sharp validates the
  decoded bytes.
- Clipboard content is rasterized to PNG in the browser when supported, with a
  server-validation fallback.
- Multer, multipart, unsupported-format, and Sharp failures return diagnostic
  error codes instead of an opaque 400.
- Stored filenames are UUID-based and user data is protected by the existing
  authentication and data-access middleware.

## Validation

Passed locally:

- `npx prettier@3.9.6 --check` on all supported changed source/docs files
- `npm test -- --run src/backend/tests/database/routes/terminal-image-utils.test.ts`
- `npm run type-check`
- targeted ESLint
- `git diff --cached --check`

The production image was built on Hoarder, transferred to the live Docker host,
and health-checked successfully. The host-visible image-path mapping was also
verified with a real uploaded PNG and a vision inspection.

The PR history was cleaned into one commit and unrelated request-origin and
homepage changes were removed.


ZacharyZcR (changes requested):
The security fundamentals here are better than most upload paths I have reviewed: decoding through sharp and re-encoding to PNG kills polyglots and SVG entirely, `limitInputPixels: 40_000_000` covers decompression bombs, the stored filename is a `randomUUID()` so nothing user-controlled reaches the filesystem, and both `authenticateJWT` and `requireDataAccess` are in place. `multer`, `sharp` and `@types/multer` are already declared, so the build is fine. The docs page is a nice touch.

Two things to fix before this can land.

## 1. The feature is broken on HTTPS deployments

`client_max_body_size 50m` was added to the server block of `docker/nginx.conf`, but `docker/nginx-https.conf` was left untouched.

`/terminal/image-upload` matches `location ~ ^/terminal(/.*)?$`, and in `docker/nginx-https.conf:313` that block sets no body-size limit — nor does anything at the `http` or `server` level in that file. So it inherits nginx's default of **1m**, and any image over 1 MB gets a 413 before it ever reaches multer. Phone screenshots clear 1 MB routinely.

The four existing `client_max_body_size 5G` entries in each file are inside other locations (`/database`, `/db`, `/ssh/file_manager/`, `/host/file_manager/ssh/`), so they do not help here and are not affected by the server-level value you added.

Adding the same line to the server block of `docker/nginx-https.conf` should be all it takes.

## 2. No rate limit on a 50 MB endpoint

An authenticated user can post 50 MB repeatedly, and `cleanupExpiredImages()` only runs on the *next* upload — so nothing reclaims space until someone uploads again, and within the one-hour TTL there is no ceiling on what accumulates. The default directory is `path.join(DATA_DIR, "termix-image-v0")`, which on a Docker install is the persistent volume holding the database.

A per-user cap, a max file count in the directory, or simply defaulting to the OS temp dir instead of `DATA_DIR` would each take the edge off. Your call which — I would not block on the specific mechanism, just on there being one.

## Minor, take or leave

- `IMAGE_FORMAT_EXTENSIONS` maps to `jpg`/`webp`/`gif`, but the write path always produces `.png`. It works — the map only gates *which* formats are accepted — but a reader will look for the branch that uses those extensions and not find it. A comment, or reducing it to a `Set` of accepted formats, would read straighter.
- `headers: { "Content-Type": undefined }` relies on axios deleting the header rather than sending it empty. Passing the `FormData` without touching headers at all is the better-supported route.

CI is approved so you can see the full suite. Ping me once the nginx side is covered.

## 3. CI is red on formatting

Now that the run is approved, `lint-and-build` fails on `prettier --check` across four files:

```
src/backend/database/routes/terminal-image-utils.ts
src/backend/database/routes/terminal.ts
src/backend/tests/database/routes/terminal-image-utils.test.ts
src/ui/features/terminal/Terminal.tsx
```

```
npx prettier@3.9.6 --write src/backend/database/routes/terminal-image-utils.ts src/backend/database/routes/terminal.ts src/backend/tests/database/routes/terminal-image-utils.test.ts src/ui/features/terminal/Terminal.tsx
```

Pin the version when you run it — the repo uses 3.9.6, and 3.8.x formats union types differently, so a global install or an editor plugin with its own bundled copy can leave files failing `--check` that look fine locally. `Postgres and MySQL` passes, so formatting is the only thing standing between this and a green run. (A pre-commit hook is on its way so this stops being your problem.)

T3rM1nAt0-R:
Addressed the review findings in commit `36f6597`:

- Added `client_max_body_size 50m` to the HTTPS nginx server block.
- Added configurable image storage caps (`TERMIX_MAX_IMAGE_COUNT`, default 100; `TERMIX_MAX_IMAGE_STORAGE_BYTES`, default 5 GiB). Expired-image cleanup, capacity checks, and writes are serialized to prevent concurrent uploads from exceeding the bounds.
- Added focused storage-limit tests.
- Reformatted the four reviewer-named files with pinned Prettier 3.9.6.

Local verification: focused tests pass (5/5), type-check passes, full lint passes with existing warnings only, production build passes, and `git diff --check` is clean.

T3rM1nAt0-R:
Also addressed the two minor review notes in the follow-up commit:

- Documented that `IMAGE_FORMAT_EXTENSIONS` is an accepted-input allowlist and that the route normalizes every upload to PNG.
- Removed the manual Axios `Content-Type` override so `FormData` gets its boundary handled normally.

Verification remains green: Prettier, focused tests (5/5), type-check, and `git diff --check`.

T3rM1nAt0-R:
Added the toolbar overlap fix in commit `$(git rev-parse --short HEAD)`: tmux detach, upload image, and paste image now share one responsive top-right action bar instead of competing absolute positions. Verified with pinned Prettier, type-check, full lint (0 errors), production build, and `git diff --check`.

T3rM1nAt0-R:
Fixed remote local-dev database connectivity in commit `$(git rev-parse --short HEAD)`. The frontend now uses the current page hostname for dev API requests instead of hardcoded `localhost`, while preserving `VITE_API_HOST` overrides. I also verified the backend with `CORS_ALLOWED_ORIGINS=http://192.168.0.21:5173`; the LAN API response now returns the matching `Access-Control-Allow-Origin` header.

T3rM1nAt0-R:
Simplified local LAN development in commit `$(git rev-parse --short HEAD)`: Vite now proxies the Termix backend service ports through same-origin `/__termix_api/<port>` paths. The browser no longer needs direct port 30001 access or a LAN CORS exception for normal dev API calls. Verified the proxied registration endpoint through port 5173 with HTTP 200.

T3rM1nAt0-R:
Re-review requested after the latest fixes.

- Deployed and verified the image-handoff build on the live Termix instance.
- Fixed the production Docker packaging issue by moving `sharp` from `devDependencies` to `dependencies` (`a5da783`).
- Rebuilt and smoke-tested the production image: backend starts successfully, `sharp` imports, container health is green, and the live listener returns HTTP 200.
- Existing data volume was preserved.

T3rM1nAt0-R:
Follow-up fix for the production 400 on both Upload image and Paste image: the UI accepts `image/*`, but the backend allowlist only accepted jpeg/png/webp/gif. Common phone formats such as HEIF/HEIC, AVIF, and TIFF were rejected after sharp decoding.

Commit `d9e485c` expands the decoded raster allowlist (SVG remains excluded), keeps normalization to PNG, adds regression coverage, and is deployed in the live container. Smoke test and production health passed; live listener returns HTTP 200.

T3rM1nAt0-R:
@ZacharyZcR Re-review requested after the PR was cleaned and rewritten into commit `5012577`.

- Removed unrelated request-origin and homepage changes.
- Retained only the terminal image handoff, storage/security limits, nginx configuration, Sharp production dependency, dev proxy, docs, and tests.
- Added clipboard PNG normalization, explicit multipart diagnostics, host-visible image path mapping, and the stable three-button terminal toolbar.
- Local Prettier, type-check, targeted ESLint, focused tests, and diff checks pass.

GitHub API does not allow the fork owner to create the formal reviewer request, so this comment is the re-review request.

ZacharyZcR (dismissed):
Both blockers are resolved, and you found a third one I had missed. Approving.

**What I asked for:**

- `docker/nginx-https.conf` now carries `client_max_body_size 50m`, so HTTPS deployments are no longer capped at nginx's 1 MB default and the feature actually works there.
- Storage is bounded, and you did it better than a rate limit would have. `TERMIX_MAX_IMAGE_COUNT` and `TERMIX_MAX_IMAGE_STORAGE_BYTES` are enforced before the write — `fileCount >= maxFileCount || totalBytes + incomingBytes > maxStorageBytes` — rather than relying on TTL cleanup to catch up. A cap on what can accumulate is the right shape for this; per-request throttling would have left the same hole open more slowly.

**What you caught yourself:** `sharp` in `devDependencies` would have broken the production Docker image, since the build installs with `--omit=dev`. Good catch — that one would have shipped and failed at runtime, not at build.

**On the widened format list:** avif, heif, jp2, jxl and tiff are all raster formats sharp decodes, and the route still runs `failOn: "error"`, `limitInputPixels: 40_000_000`, a `metadata()` format check, and re-encodes everything to PNG. Critically **svg is still absent** — that is the one that would have mattered, since it is the polyglot and SSRF-via-external-entity vector. Accepting phone formats without loosening the security posture is exactly right.

## One thing for next time, not blocking

`vite.config.ts` (12 proxied ports) and the `main-axios.ts` dev-host change are local-development improvements with nothing to do with terminal image handoff. I checked their blast radius before deciding not to hold the PR for it: `__termix_api` is inside `else if (devMode)`, and `apiHost` is only read at line 724 in that same branch — production takes `getBasePath() + path` and is untouched. So it is zero risk to ship.

But it is still scope that arrived after you had already stripped unrelated changes out once. Separate PRs make each easier to review, and easier to revert if one turns out wrong. Worth splitting next time rather than folding a second improvement into a feature branch.

## Merging

CI is approved and running. Once both checks are green this is good to go from my side — the security fundamentals were solid from the first revision, and the operational gaps are now closed.

Thanks for the three rounds, and for the deployment testing against a live instance. The HEIF report in particular is the kind of thing that only shows up in real use.

PR #1171, opened 2026-08-08
Make audit_logs.user_id nullable so the trail can outlive the account

ZacharyZcR:
Fixes Termix-SSH/Support#1097.

## The defect

Deleting a user fails on any account that has ever logged in, because `AuditLogRepository.anonymizeByUserId` tries to null `user_id`:

```
SqliteError: NOT NULL constraint failed: audit_logs.user_id
    at AuditLogRepository.anonymizeByUserId (audit-log-repository.js:90)
    at deleteUserAndRelatedData (delete-user-data.js:15)
```

Two independent call paths hit it: the admin user deletion in `users.js`, and the cleanup step of `POST /users/link-oidc-to-password`.

## Root cause

`src/backend/database/db/index.ts` had two conflicting definitions of `audit_logs`:

- the primary bootstrap (~line 508) created `user_id TEXT NOT NULL`;
- a second `CREATE TABLE IF NOT EXISTS` inside `migrateSchema()` (~line 1732) had the column correctly nullable — but it can never run. The table already exists by then, so `IF NOT EXISTS` is a no-op.

Everything else in the codebase already agreed that the column is nullable: `schema.ts` documents it as deliberate ("the trail outlives the account, and username keeps the entry attributable"), `drizzle/sqlite/0000_clever_hercules.sql` declares `user_id text` with `ON DELETE set null`, and the repository was written against that. Only the runtime bootstrap disagreed — which is why this hits **fresh installs**, not just migrated databases.

## The fix

- Align the primary bootstrap with the schema: `user_id TEXT` with `ON DELETE SET NULL`.
- Rebuild the table on existing databases, using the same create-copy-drop-rename pattern already used in `migrateSchema()` for `ssh_credentials.username` — SQLite has no `ALTER COLUMN`. The copy lists columns explicitly, and rows are preserved.

I left the unreachable second definition alone since it now matches the primary one and no longer conflicts; happy to remove it in a follow-up if you would rather not keep dead code around.

## Tests

`src/backend/tests/database/db/audit-log-user-id-nullable.test.ts` boots the real database twice:

- **fresh install** — insert a user and an audit entry, delete the user, assert `user_id` is `NULL` and `username` survives. This is the exact path that used to throw.
- **existing database** — seed a `db.sqlite` whose `audit_logs.user_id` still has the constraint plus one row, boot, assert the column is nullable and the row is still there.

Both fail on `dev-2.7.0` without the fix and pass with it. Full suite: 141 files, 1055 tests passing. `tsc --noEmit`, `npm run lint` (0 errors) and Prettier are clean.

### DIFF
--- a/src/backend/database/db/index.ts
+++ b/src/backend/database/db/index.ts
@@ -507,7 +507,7 @@ async function initializeCompleteDatabase(): Promise<void> {
 
     CREATE TABLE IF NOT EXISTS audit_logs (
         id INTEGER PRIMARY KEY AUTOINCREMENT,
-        user_id TEXT NOT NULL,
+        user_id TEXT,
         username TEXT NOT NULL,
         action TEXT NOT NULL,
         resource_type TEXT NOT NULL,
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
The **Postgres and MySQL** check here fails for a reason unrelated to this PR: `resyncAutoIncrement` in the repository test support calls `pg_get_serial_sequence(table, 'id')` on every seeded table, and that raises 42703 rather than returning null for `host_sidebar_preferences`, which is keyed on `user_id` and has no `id` column. It fails the same way for every pull request against `dev-2.7.0`.

#1173 fixes it, and its own Postgres job passes. Once that lands, a re-run here should go green — this branch only touches `db/index.ts`.

PR #1172, opened 2026-08-08
Derive the ssh_credentials rebuild from the live schema instead of a stale column list

ZacharyZcR:
Noticed while working on #1171, which reuses this rebuild pattern. No issue filed for it — it fails silently, so nobody would report it as this.

## The defect

`migrateSchema()` rebuilds `ssh_credentials` on startup to drop the old `username NOT NULL` constraint, since SQLite cannot `ALTER` a column. That rebuild has never been able to complete.

It restated the table as a literal with 19 columns and then copied rows positionally:

```js
INSERT INTO ssh_credentials_temp_migration SELECT ${allColumns} FROM ssh_credentials;
```

where `allColumns` is *every* live column. But the table has gained columns since that literal was written, and all of them are added by `addColumnIfNotExists` immediately **before** the rebuild runs:

- `cert_public_key`
- `pin`
- `sort_order`
- `sync_id`

So the destination is narrower than the source. SQLite rejects the INSERT with a column count mismatch, the `catch` downgrades it to `databaseLogger.warn`, and the constraint is still there on the next boot — every boot, forever. Databases upgraded from before key-only credentials therefore still refuse a credential with no username.

The positional copy is the second half of the trap: had the counts happened to line up after a future column was added, the mismatch would have become a silent column shift instead of an error.

And `DROP TABLE` discards the table's indexes. `idx_ssh_credentials_sync_id` is `UNIQUE` and identifies rows to remote sync; it is only recreated by a `CREATE UNIQUE INDEX IF NOT EXISTS` about 1400 lines further down the same function. Correct today, but only by ordering.

## The fix

- Read the `CREATE TABLE` statement back from `sqlite_master` and rewrite only the table name and the `username` constraint. The replacement table is derived from the real one, so it cannot fall behind it — including for columns added after this code.
- Throw (rather than quietly rebuild an identical table) if the constraint rewrite does not match anything.
- Copy rows by explicit column name, not positionally.
- Replay the table's index definitions after the rename instead of relying on later code to notice.

## Tests

`src/backend/tests/database/db/ssh-credentials-username-rebuild.test.ts` seeds a `db.sqlite` in exactly the state a 2.6.x run leaves behind — old constraint, all the newer columns, the unique `sync_id` index, one row — then boots the real database and asserts:

- the constraint is gone and a key-only credential inserts;
- every column survives, `sync_id` included (losing it would re-key the record for remote sync);
- `idx_ssh_credentials_sync_id` still exists and still rejects a duplicate.

The first and third fail on `dev-2.7.0` today. Full suite: 140 files, 1056 tests passing. `tsc --noEmit`, `npm run lint` (0 errors) and Prettier are clean.

Independent of #1171 — that one targets `audit_logs` and only touches a different block of this file.

### DIFF
--- a/src/backend/database/db/index.ts
+++ b/src/backend/database/db/index.ts
@@ -1117,43 +1117,57 @@ const migrateSchema = () => {
 
     if (usernameCol && usernameCol.notnull === 1) {
       const tempTableName = "ssh_credentials_temp_migration";
-      const allColumns = tableInfo.map((col) => col.name).join(", ");
+      const allColumns = tableInfo.map((col) => `"${col.name}"`).join(", ");
+
+      // Derive the replacement table from the live definition instead of
+      // restating it here. The table keeps gaining columns (cert_public_key,
+      // pin, sort_order, sync_id, ...), and a second copy of the column list
+      // falls behind every time one is added — leaving the copy narrower than
+      // the table, so the INSERT below fails and the constraint stays put.
+      const createSql = sqlite
+        .prepare("SELECT sql FROM sqlite_master WHERE type = 'table' AND name = 'ssh_credentials'")
+        .pluck()
+        .get() as string | undefined;
+
+      if (!createSql) {
+        throw new Error("ssh_credentials has no stored table definition");
+      }
+
+      // Only the table name is rewritten; replace() stops at the first match,
+      // and in a CREATE TABLE statement that is the table being defined.
+      const renamedSql = createSql.replace("ssh_credentials", tempTableName);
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
The **Postgres and MySQL** check here fails for a reason unrelated to this PR: `resyncAutoIncrement` in the repository test support calls `pg_get_serial_sequence(table, 'id')` on every seeded table, and that raises 42703 rather than returning null for `host_sidebar_preferences`, which is keyed on `user_id` and has no `id` column. It fails the same way for every pull request against `dev-2.7.0`.

#1173 fixes it, and its own Postgres job passes. Once that lands, a re-run here should go green — this branch only touches `db/index.ts`.

PR #1183, opened 2026-08-09
fix: use getClientIp in getRequestMeta for correct audit-log IPs

brennanneoh:
## Summary
Follow-up to #1169, per review comment: `getRequestMeta` in `src/backend/utils/audit-logger.ts` had near-identical forwarded-header logic to `getClientIp` but strictly worse — its array branch didn't split/trim, it never fell back to the socket peer, and it returned `""` instead of `"unknown"`. Now delegates to `getClientIp` so the audit trail gets the same correctness as the terminal login-alert path.

Note: this branch was cut from `worktree-snazzy-hugging-parrot` (source branch of #1169) before that branch was deleted on merge, so this diff still carries the `getClientIp` commits in its history. If `main` already has them by the time this is reviewed, that history should collapse cleanly; otherwise this PR is self-contained.

## Test plan
- [x] `npx vitest run src/backend/tests/utils/audit-logger.test.ts src/backend/tests/utils/request-origin.test.ts` — 24 passed
- [x] `npx tsc --noEmit`

### DIFF
--- a/src/backend/hosts/terminal/index.ts
+++ b/src/backend/hosts/terminal/index.ts
@@ -45,6 +45,7 @@ import {
 import { isWindowsSftpPath, sftpPathToLocalPath } from "../transfer-paths.js";
 import { preparePrivateKeyForSSH2 } from "../../utils/ssh-key-utils.js";
 import { triggerLoginAlert } from "../../utils/alert-trigger.js";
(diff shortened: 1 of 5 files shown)


LukeGus:
LGTM, thanks.

PR #1198, opened 2026-08-11
fix: make mobile terminal scrollback match xterm wheel behavior

T3rM1nAt0-R:
## Why

Mobile scrolling in the web app was frustrating: single-finger gestures did
not behave like normal terminal wheel scrolling, and regular shell scrollback
could behave differently from alternate-screen terminal applications. This
change makes mobile touch scrolling use xterm's own wheel and scrollback paths
so the interaction is consistent with desktop terminal behavior.

## Summary

- Route one-finger mobile terminal drags through xterm's native wheel path.
- Dispatch synthetic wheel events at `.xterm-scrollable-element` so normal
  shell scrollback uses xterm's real viewport handler.
- Preserve alternate-buffer/TUI and tmux wheel behavior through xterm's
  existing terminal handler.
- Add bounded post-release kinetic momentum with cancellation and reduced-motion
  support.
- Add unit tests plus real-xterm normal-buffer and alternate-buffer regression
  coverage.

## Root cause

Synthetic touch wheel events were dispatched at `.xterm-screen`. That reached
xterm's alternate-buffer/TUI wheel handling, but did not reliably enter the
normal scrollback viewport seam. The normal shell path is owned by
`.xterm-scrollable-element`.

## Verification

- `npm test`: 206 test files passed; 1,486 tests passed; 1 skipped.
- Focused touch/xterm tests: 18/18 passed.
- Type-check passed.
- Prettier and `git diff --check` passed.
- Production Docker build passed.
- Isolated container smoke test passed.
- Manual mobile acceptance passed for regular terminal scrollback and Hermes
  TUI behavior.

## Scope

No history-overlay code, runtime changes, generated output, credentials, or
unrelated files are included.

Momentum speed tuning is intentionally left as a separate follow-up.

## Documentation

- Added `docs/terminal-touch-input.md` covering direct touch scrolling, bounded release momentum, normal shell scrollback, tmux/alternate-screen behavior, cancellation, and reduced-motion behavior.

### DIFF
--- a/docs/terminal-touch-input.md
+++ b/docs/terminal-touch-input.md
@@ -1,0 +1,54 @@
+# Mobile Touch Scrolling
+
+Termix terminals support one-finger touch scrolling on mobile devices. Touch
+scrolling uses the terminal's normal wheel-input path so it works with both
+regular shell scrollback and full-screen terminal applications.
+
+## Scrolling a terminal
+
+- Drag one finger vertically across the terminal to scroll directly.
+- Release a fast swipe to continue with bounded momentum.
+- Drag upward to move toward earlier output.
+- Drag downward to move toward newer output.
+
+The behavior follows the terminal's existing xterm wheel handling. This keeps
+normal shell scrollback, tmux, alternate-screen applications, and terminal
+mouse-reporting behavior aligned with desktop wheel input.
+
+## Interaction behavior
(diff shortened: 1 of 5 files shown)


LukeGus:
LGTM, thanks.

PR #1232, opened 2026-08-14
feat: tune host transfers adaptively

ZacharyZcR:
## Summary

- make automatic lane selection the default while preserving explicit overrides
- tune SFTP pipeline depth and lane count from file size plus a bounded host-pair EWMA profile
- back off concurrency after failures and expire learned profiles after 24 hours
- sample at most three 64 KiB file regions to improve tar routing beyond filename extensions
- keep SHA-256 verification and existing relay/direct fallback behavior unchanged

Closes Termix-SSH/Support#1150

## Validation

- `npm run lint` (0 errors; existing warnings only)
- `npm run type-check`
- `npm run build`
- targeted transfer tests: 13 passed
- full suite: 2262 passed, 1 skipped; one unrelated session-log test timed out under full-suite load and passed twice when rerun in isolation


ZacharyZcR:
Tracking issue: Termix-SSH/Support#1150.

PR #1243, opened 2026-08-14
Fix .preferred_username when using LDAP login.

njz-cvm:
# Overview

_Termix allows hosts to login with a user's oidc username. This works for normal OIDC, but not with LDAP. This PR seeks to fix that._

- [X] Added: Tests to cover the changes made in this PR.
- [ ] Updated: N/A
- [ ] Removed: N/A
- [X] Fixed: $oidc.preferred_username behaves as expected when using LDAP.

# Changes Made

When expanding a oidc username, these additional checks are made:
 - If the oidc username matches the LDAP prefix regex...
 - And the user's SSO directory is LDAP...
 - And the user's directory ID matches the oidc username prefix ID...
 - Use the LDAP uid for oidc username.

# Checklist

- [X] Code follows project style guidelines
- [X] Supports mobile and desktop UI/app (if applicable)
- [X] I have read [Contributing.md](https://github.com/Termix-SSH/Termix/blob/main/CONTRIBUTING.md)
- [X] This is not a translation request. See [docs](https://docs.termix.site/translations)


LukeGus:
LGTM, thanks.

PR #1347, opened 2026-08-25
fix: stop SSH authentication during routine status polling

### LINKED ISSUE #1216: [BUG]
### Title

Status check triggers a full SSH authentication attempt even when metricsEnabled is false

### Platform

Desktop App - Windows

N/A

### Version

2.7.1

### CLI Installation Method

None

### CLI Version

_No response_

### Troubleshooting

- [x] I have checked open issues and ensured this is not a duplicate

### The Problem

Status check on an SSH host with metrics disabled is supposed to just be a TCP ping. Instead it's doing a full SSH auth attempt on every status check interval. On a RADIUS/Duo backed switch this means a live Duo push every interval, for every host, with zero tabs or sessions open. 

### How to Reproduce

1. Install Termix 2.6.0 (or earlier), add SSH hosts with statusCheckEnabled: true and metricsEnabled: false, confirm status checks are just TCP pings (no auth attempts on the target).
2. Upgrade to 2.7.1 via the MSI installer
3. Open the app
4. Watch the auth logs on an SSH host that has statusCheckEnabled: true and metricsEnabled: false a full SSH login attempt fires on that host, with no tabs open.
5. To rule out a stale config: uninstall, delete the app data folder, reinstall fresh.
6. Import hosts from a JSON export (statsConfig included, metricsEnabled: false on each host).
7. As soon as the import finishes, the same behavior happens. SSH login attempts fire against every imported host on the next status check interval, still with no tabs open.

### Additional Context

Easier to see with switches where radius auth 2fa enabled. Otherwise you'll have to look through local appdata to see session logs.

YatoVoid:
Routine host status polling is now TCP-only, preventing repeated RADIUS/Duo authentication prompts when no SSH session is active. Metrics collection remains responsible for authenticated `online` status, and regression coverage ensures `pollHostStatus` cannot call the SSH authentication path.

Fixes Termix-SSH/Support#1216

### DIFF
--- a/src/backend/hosts/metrics/host-status.test.ts
+++ b/src/backend/hosts/metrics/host-status.test.ts
@@ -1,6 +1,5 @@
 import { describe, expect, it } from "vitest";
 import {
-  needsStatusPollAuthentication,
   statusAfterAuthentication,
   statusAfterReachabilityCheck,
 } from "./host-status.js";
@@ -21,12 +20,5 @@ describe("host availability status", () => {
   it("downgrades failed authentication without hiding reachability", () => {
     expect(statusAfterAuthentication(false, "online")).toBe("reachable");
     expect(statusAfterAuthentication(false, "offline")).toBe("offline");
-  });
-
-  it("authenticates in the status probe unless a metrics poll will do it", () => {
-    expect(needsStatusPollAuthentication(false, false)).toBe(true);
-    expect(needsStatusPollAuthentication(false, true)).toBe(true);
-    expect(needsStatusPollAuthentication(true, false)).toBe(true);
-    expect(needsStatusPollAuthentication(true, true)).toBe(false);
   });
 });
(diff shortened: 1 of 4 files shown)


ZacharyZcR (changes requested):
The current guard only stops polling authentication after a successful authentication has already moved the host to `online`. If the Duo/RADIUS request is denied, ignored, or times out, `statusAfterAuthentication(false)` stores `reachable`; `shouldProbeAuthForStatus("reachable")` then returns `true` on the next interval, so the repeated 2FA push reported in Support#1216 continues indefinitely.

Please change the metrics-disabled status path so routine status polling does not perform SSH authentication (the issue explicitly expects TCP-only polling), or otherwise ensure failed/interactive authentication is not retried by each status interval. Please also add a polling-level regression test covering authentication failure followed by the next poll; the current helper-only tests encode the retry behavior and cannot catch this failure mode.

ZacharyZcR (approved):
The requested behavior and regression coverage are now addressed. Routine status polling is TCP-only, repeated interactive SSH authentication is removed, and the status semantics are documented accurately.

PR #1374, opened 2026-08-31
fix(homepage): validate clock widget timezones

nasif-naseef:
# Overview

- [x] Fixed: an invalid timezone on a Clock widget threw `RangeError` during render and took the homepage canvas down with it
- [x] Added: timezone validation in the Clock edit dialog, with an inline message and a disabled Save
- [x] Updated: `ClockWidget` now falls back to local time for configs already stored with an unusable zone

# Changes Made

- `src/ui/features/homepage/clock-timezone.ts` (new) — `normalizeTimezone`, `isValidTimezone` and `validateClockTimezone`, shared by the widget and the dialog so the check lives in one place.
- `ClockEditForm.tsx` — flags an unusable zone the same way `FolderMetadataDialog` flags a duplicate folder name: `aria-invalid`, `border-destructive`, and an inline `text-destructive` message.
- `WidgetEditDialog.tsx` — one `typeId === "clock"` branch: Save is disabled while the zone is unusable, `handleSave` guards again, and the normalized zone is what gets stored, so `America/New York` is saved as `America/New_York` instead of being rejected. No other widget form is touched.
- `ClockWidget.tsx` — validates before formatting, falling back to local time (and hiding the zone label) for configs saved before this change, so an existing broken homepage recovers on its own.
- `en.json` — one new key, `homepage.invalidTimezone`.
- Tests — `ClockTimezone.test.tsx` and `WidgetEditDialogClock.test.tsx`, 38 assertions covering the widget fallback, the normalizer, the inline error, and Save being blocked and then re-enabled.

# Related Issues

- Related to Termix-SSH/Support#1238

# Screenshots / Demos

Invalid zone: the field is flagged and Save is disabled, so it can no longer be stored.

invalid zone, error shown, Save disabled

A space instead of an underscore is accepted and normalized on save — no error, Save enabled.

space accepted, Save enabled

Saved result — the widget renders the zone as `America/New_York`:

widget rendering America/New_York

Before this change, that same config threw during render and took the canvas down with it:

```
RangeError: Invalid time zone specified: America/New York
```

# Checklist

### DIFF
--- a/src/ui/features/homepage/clock-timezone.ts
+++ b/src/ui/features/homepage/clock-timezone.ts
@@ -1,0 +1,28 @@
+/** IANA zone names use underscores, but people type and paste spaces. */
+export function normalizeTimezone(timezone: string): string {
+  return timezone.trim().replace(/\s+/g, "_");
+}
+
+export function isValidTimezone(timezone: string): boolean {
+  try {
+    new Intl.DateTimeFormat("en-US", { timeZone: timezone });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Normalizes the timezone a user typed and reports whether it is usable.
+ * A blank field is valid: it means "use local time".
+ */
+export function validateClockTimezone(timezone: unknown): {
+  timezone?: string;
+  valid: boolean;
+} {
+  if (typeof timezone !== "string" || timezone.trim() === "") {
+    return { timezone: undefined, valid: true };
+  }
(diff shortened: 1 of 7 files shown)


ZacharyZcR (approved):
Reviewed the timezone validation, persisted-config fallback, and test coverage. The focused tests, full frontend suite, type-check, lint, formatting, and diff checks all pass. No blocking changes requested.

PR #1382, opened 2026-09-03
fix(desktop): surface a failed embedded backend start

nasif-naseef:
# Overview

The desktop app hung forever on the "Loading…" spinner when the embedded backend's HTTP port was already in use, with nothing in the UI saying why. It now reports the failure and names the conflicting port.

- [x] Added: `electron/backend-failure.cjs` — classifies why the forked backend died (port conflict vs. generic crash) from its exit code and stderr
- [x] Added: `src/ui/lib/embedded-server-status.ts` — renderer-side read of that verdict over IPC
- [x] Fixed: the three renderer retry loops no longer wait forever on a backend that has already exited
- [x] Updated: `get-embedded-server-status` now carries the failure reason, and is finally reachable from the renderer

# Changes Made

The backend already handled the conflict correctly — `httpServer.on("error", …)` in `src/backend/database/database.ts` logs it and exits 1 — and the main process noticed, with `startBackendServer` resolving `false`. But that verdict had nowhere to go: `get-embedded-server-status` was never exposed in `electron/preload.js` and nothing in `src/` called it, so the renderer never learned the backend was gone.

The renderer treats every connection failure as "the embedded backend is still booting" and retries indefinitely, in three places, each with a comment explaining why giving up would be wrong:

- `src/main.tsx`, the `phase === "verifying"` effect — the loop that actually produced the observed permanent spinner, since a launch with a stored session is the common case
- `src/ui/auth/Auth.tsx`, the desktop auto-session effect — a login form is not a valid destination for a standalone install whose only local user has no password
- `src/ui/features/FullScreenAppWrapper.tsx` — same reasoning for popped-out windows

That assumption is right while the backend is slow to boot and wrong once it has exited, and nothing distinguished the two. So rather than capping those retries — which would reintroduce the "spurious logout / login form for a passwordless local user" problems the comments warn about — each loop now keeps waiting only while the main process reports *no* failure, and hands over to the error screen once one is reported.

`electron/main.cjs` keeps a bounded (8 KB) tail of the child's stderr so a failure can be classified after the fact, classifies on `exit` whether or not the 15s ready promise already settled optimistically, and suppresses classification while we are deliberately tearing the backend down on quit. A port conflict reports the port, since that is the part the user can act on — most often a Termix container on the same ports, or a backend orphaned by a hard-kill that `reapOrphanedBackendProcess` could not claim because the holder was not our own recorded PID.

Secondary service ports (`30003`–`30012`) turned out not to be affected in the same way: with one of those occupied the backend still reaches `backend_init_complete` and the app loads, so this change is scoped to a backend that actually exits.

# Related Issues

- Closes Termix-SSH/Support#1254

# Screenshots / Demos

Reproduced and verified end-to-end against a real `linux-unpacked` build, with `30001` held by another process.

**Before** — main process knew, renderer never did, spinner never resolved:

```
[backend:stderr] Error: listen EADDRINUSE: address already in use :::30001
  code: 'EADDRINUSE', port: 30001
Backend process exited with code 1, signal null
startBackendServer result: false
=== Startup complete ===
```
Rendered UI text, indefinitely: `Loading...`

**After** — same launch, `~/.config/termix/termix-main.log`:

```
[backend:stderr] Error: listen EADDRINUSE: address already in use 127.0.0.1:30001
  code: 'EADDRINUSE',
Backend failure classified as: port-in-use
startBackendServer result: false
```

Rendered UI text, read out of the running app over CDP:

```
Termix's local server could not start

Port 30001 is already in use by another program. This is usually a Termix
Docker container or a leftover Termix process from a previous session.
Close it, then restart Termix.
```

Regression checks on the same build: with the port free the app starts normally (`backend_init_complete`, `startBackendServer result: true`, full UI renders) and quitting reports no false crash.

New tests: 18 across `scripts/electron-backend-failure.test.ts` (classification, including the real captured stderr, a truncated tail, a signal kill, and a deliberate shutdown), `src/ui/tests/lib/embedded-server-status.test.ts` (still-booting, both failure kinds, an older build with no channel, IPC rejection, unknown reason), and `src/ui/tests/auth/AuthEmbeddedServerFailure.test.tsx` (the rendered screen for each case, and that a slow boot still shows no error).

`npx vitest run` goes from 2682 to 2700 passing with no new failures; the 119 pre-existing failures on `dev-2.8.0` are unrelated (jsdom in this project has no `localStorage` global, which several existing suites call directly). `npm run lint` reports 0 errors, `tsc -b` and `prettier --check` are clean.

# Checklist

### DIFF
--- a/electron/backend-failure.cjs
+++ b/electron/backend-failure.cjs
@@ -1,0 +1,45 @@
+// The embedded backend is a forked child process, so the only evidence the
+// main process gets when it dies is an exit code plus whatever it wrote to
+// stderr on the way out. A port conflict is by far the most common cause
+// (a leftover backend from a hard-killed session, a Termix container bound
+// to the same ports, or an unrelated service on 30001) and it is also the
+// only one the user can actually act on, so it is classified separately
+// from a generic crash and reported with the port that was taken.
+
+const SIGNALS_MEANING_DELIBERATE_SHUTDOWN = new Set(["SIGTERM", "SIGINT"]);
+
+function parseConflictingPort(stderr) {
+  // Node's EADDRINUSE error prints the port twice: once inside the message
+  // ("address already in use :::30001") and once as a numeric `port` field
+  // on the error object. Either is enough, but stderr arrives in chunks and
+  // the tail can be cut off, so both are tried.
+  const fromErrorField = stderr.match(/\bport:\s*(\d{1,5})\b/);
+  if (fromErrorField) return Number(fromErrorField[1]);
+
+  const fromMessage = stderr.match(/address already in use\s+\S*?:(\d{1,5})\b/);
+  if (fromMessage) return Number(fromMessage[1]);
+
+  return null;
+}
+
+/**
+ * Classifies why the embedded backend process ended.
+ *
+ * Returns null when nothing went wrong -- a clean exit, or a shutdown we
+ * asked for -- and otherwise the reason the renderer should surface instead
+ * of waiting forever for a backend that is never coming back.
+ */
+function classifyBackendFailure({ exitCode, signal = null, stderr = "" }) {
+  if (exitCode === 0) return null;
+  if (exitCode === null && SIGNALS_MEANING_DELIBERATE_SHUTDOWN.has(signal)) {
+    return null;
+  }
+
+  if (stderr.includes("EADDRINUSE")) {
+    return { reason: "port-in-use", port: parseConflictingPort(stderr) };
+  }
+
+  return { reason: "crashed", port: null };
+}
+
+module.exports = { classifyBackendFailure };
--- a/electron/backend-failure.cjs
+++ b/electron/backend-failure.cjs
@@ -1,12 +1,20 @@
 // The embedded backend is a forked child process, so the only evidence the
-// main process gets when it dies is an exit code plus whatever it wrote to
-// stderr on the way out. A port conflict is by far the most common cause
-// (a leftover backend from a hard-killed session, a Termix container bound
-// to the same ports, or an unrelated service on 30001) and it is also the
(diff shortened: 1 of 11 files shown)


ZacharyZcR (changes requested):
The overall approach is sound, but the exit classifier currently leaves a permanent-loading path open. `backendStopRequested` already suppresses classification for shutdowns initiated by Termix. Once that guard has passed, *any* backend exit means the renderer can no longer recover by retrying.

At present `classifyBackendFailure()` returns `null` for exit code 0 and for SIGTERM/SIGINT. An embedded backend that exits cleanly on its own, is terminated by the OS/service manager, or receives an external SIGTERM will therefore set `backendProcess = null` with `backendFailure = null`; all three renderer loops will resume waiting forever—the exact failure mode this PR is intended to eliminate. The test currently codifies that behavior as “deliberate shutdown,” but the classifier has no evidence that the shutdown was deliberate; that evidence exists only in `backendStopRequested`.

Please remove the clean-exit and signal exemptions from the classifier (or pass an explicit `stopRequested` flag into it), and update the tests so every exit not initiated by `stopBackendServer()` becomes `crashed`. Keep `backendStopRequested` as the sole deliberate-shutdown guard in the main process. After that change and green required CI, this should be reviewable.

ZacharyZcR (approved):
Re-reviewed at 73ba88047e4e9de2157ff1de673216d9f454c5cf. The previous blocker is resolved: backendStopRequested is now the sole deliberate-shutdown guard, and every unrequested child exit is surfaced as a terminal failure. Local verification passed: TypeScript project check and 16 focused regression tests across backend classification, renderer status handling, and Auth failure UI.

PR #1385, opened 2026-09-03
fix(guacamole): coalesce VNC wheel floods and label meta keys by device

neooriginal:
## Summary
- Cap guacamole-common-js wheel-button floods so VNC scrolling stays usable over high-latency proxies such as Cloudflare, instead of queuing seconds of leftover scroll. The websocket path is unchanged.
- Label the VNC toolbar Super/Cmd on non-Windows clients instead of always showing Windows key / Win+L. RDP stays Windows-branded.

## Test plan
- [x] Open a VNC session through a Cloudflare (or similarly lagged) remote URL and scroll with a mouse wheel and a trackpad. Scrolling should move promptly and stop when you stop, not crawl for a long time afterward.
- [x] Confirm a local/low-latency VNC session still scrolls by the expected amount for a single wheel notch.
- [x] On macOS, open a VNC session and confirm the toolbar shows Cmd / Cmd+L, not Win.
- [x] On Linux, open a VNC session and confirm Super / Super+L.
- [x] On Windows or any RDP session, confirm Win / Win+L are unchanged.
- [x] Sticky modifier still sends Super_L (0xffeb) after the label change.


ZacharyZcR (approved):
Reviewed at d44a7b37bf4f111e4290dc62fc566ea4a0a9f721. The scroll coalescer preserves Guacamole press/release pairs, caps queued wheel clicks, clears pending timers on teardown, and leaves the remote keysym unchanged while adjusting only the device-facing label. Local verification passed: TypeScript project check and all 15 focused Guacamole tests.

PR #1386, opened 2026-09-03
Add white label branding settings

neooriginal:
## Summary
- Admins can set a custom application name, logo and tagline from Admin Settings.
- Values are applied on the login screen and to the browser tab title/favicon at runtime.
- Stored in the existing settings table, no migration required. Logo uploads are validated and re-encoded to PNG (max 750KB), with GET public (needed pre-login) and PATCH admin-only.

## Test plan
- [x] `tsc -b --force` passes
- [x] `vite build` passes
- [x] Backend route test suite passes (399/399)
- [x] Manual check: `GET /users/branding` returns defaults publicly, `PATCH /users/branding` requires auth
- [x] Manual check in Admin Settings: set name/logo/tagline, verify login screen, tab title and favicon update


ZacharyZcR (changes requested):
The implementation direction looks reasonable, but this is not safe to merge yet because the new backend trust boundary has no regression coverage. The only added tests cover the browser file pre-check and document icon helper; none exercise parseBrandingLogoDataUrl() or the public/admin routes. Please add backend tests that prove: unauthenticated GET succeeds; unauthenticated and non-admin PATCH are rejected; admin PATCH persists only allowed fields; malformed/non-raster/oversized image data is rejected; and a valid image is decoded and normalized. The existing 399 route tests passing cannot catch regressions in routes and validation that those tests never invoke. Please also run the repository-required CI on the updated head.

ZacharyZcR (approved):
Re-reviewed at 1cdcb802e86d0b24aed81c8ae1ce54b7f5aef740. The requested backend coverage is now present and verifies public GET, admin-only PATCH, field validation, malformed/oversized input rejection, and valid image normalization. Local verification passed: 29 focused tests and full TypeScript project check.

PR #1388, opened 2026-09-04
fix(backend): report port conflicts on the service ports

nasif-naseef:
# Overview

A port conflict on any of the backend's secondary service ports (`30003`–`30012`) started nothing and reported nothing. The backend logged a healthy start, the app loaded, and the affected feature was silently dead for the session.

- [x] Added: `src/backend/utils/service-listen.ts` — `listenOnServicePort()` and `attachServicePortConflictHandler()`
- [x] Fixed: all seven fixed-port services now detect and report a conflict instead of pretending to start
- [x] Added: unit tests plus a source-level guard so no service reaches for `app.listen()` again

# Changes Made

Express's `app.listen(port, host, callback)` registers the caller's callback as the server's `error` handler as well as its `listening` handler:

```js
// express/lib/application.js
app.listen = function listen() {
  var server = http.createServer(this)
  var args = slice.call(arguments)
  if (typeof args[args.length - 1] === 'function') {
    var done = args[args.length - 1] = once(args[args.length - 1])
    server.once('error', done)      // <-- the "started" callback
  }
  return server.listen.apply(server, args)
}
```

So a bind failure calls the success callback with the error as its first argument and emits nothing:

```
$ node -e 'require("express")().listen(30005,"127.0.0.1",(...a)=>console.log(a.map(x=>x&&x.code)))'
[ 'EADDRINUSE' ]
```

Every service wrote that callback as `() => {…}` and ignored its arguments, so a failed bind was indistinguishable from a successful one. This also explains why only `30001` ever behaved sensibly — `database.ts` does not use `app.listen()`; it builds the server and attaches a real `error` handler.

Worth calling out explicitly: **adding `.on("error")` to the services would not have fixed this.** Express consumes the event through `once('error', done)` before any later handler could run. The callback has to stop being passed to `listen()` at all, which is what the new helper does.

`listenOnServicePort()` builds the server so `listening` and `error` stay separate, and handles a conflict the way `database.ts` already does — name the port, then exit — rather than running on with one feature missing. Exiting is also what lets the desktop tell the user which port to free, via the classifier added in #1382. The tunnel service owns its own `http.Server` for the WebSocket upgrade path, so it takes the same handler through `attachServicePortConflictHandler()`.

Converted: tunnel `30003`, file-manager `30004`, metrics `30005`, dashboard `30006`, docker `30007`, tmux `30010`, homepage `30012`.

**Deliberately out of scope:** the WebSocket services on `30002`, `30008`, `30009` and `30011` use `new WebSocketServer({ host, port })` rather than `app.listen()`. They are not affected by this express behaviour and do emit an error, so they are left alone here; happy to fold them in if you would rather have one uniform policy.

Behaviour note: `file-manager` previously logged a conflict via a handler registered *after* `app.listen()` but kept running. It now exits like the others. Its existing runtime `server.on("error")` logging and its `requestTimeout = 0` are unchanged.

# Related Issues

- Fixes Termix-SSH/Support#1260 — cross-repo, so GitHub's closing keywords do not apply here; the issue needs closing by hand on merge, the same way Support#1254 was handled for #1382.
- Follow-up to #1382 / Termix-SSH/Support#1254, which fixed the main-port case.

# Screenshots / Demos

Verified on a real `linux-unpacked` build with `127.0.0.1:30005` held by another process.

**Before** — backend reports success, app loads, metrics silently dead, and `grep -icE "eaddrinuse|uncaught|30005"` across the backend log, the Electron main log and stdout returns `0`:

```
[SUCCESS] [🚀] Termix backend started successfully [op:backend_init_complete,duration:927ms]
```

```
127.0.0.1:30001  termix          127.0.0.1:30005  ← the squatter. termix never got it.
…                                127.0.0.1:30006  termix
```

**After** — same build, same conflict:

```
[backend:stderr] [ERROR] [📊] Port 30005 is already in use. Kill the existing process and retry. [op:service_port_conflict]
Error: listen EADDRINUSE: address already in use 127.0.0.1:30005
Backend process exited with code 1, signal null
Backend failure classified as: port-in-use
```

and the desktop shows, naming the real port:

> **Termix's local server could not start**
> Port 30005 is already in use by another program. This is usually a Termix Docker container or a leftover Termix process from a previous session. Close it, then restart Termix.

**Regression, all ports free:** all 12 service ports bind, `0` conflicts logged, UI renders normally, and `curl 127.0.0.1:30005` returns `401` — the metrics service is up and authenticating.

Tests: 5 new. Four cover the helper (starts and reports listening on a free port; does **not** invoke the ready callback when the port is taken; logs the port and `operation: "service_port_conflict"`; leaves non-`EADDRINUSE` errors to the process handlers). The fifth scans `src/backend` and fails if any file reintroduces `app.listen(`.

`npx vitest run` goes from 2708 to 2713 passing with no new failures — the 119 pre-existing failures on `dev-2.8.0` are unrelated and I confirmed that baseline by stashing. `tsc -b` clean, `npm run lint` 0 errors with the warning count unchanged at 102, `prettier --check` clean.

# Checklist

### DIFF
--- a/src/backend/hosts/docker/index.ts
+++ b/src/backend/hosts/docker/index.ts
@@ -16,6 +16,7 @@ import {
   getRequestUserId,
   registerDockerSshRoutes,
 } from "./routes.js";
+import { listenOnServicePort } from "../../utils/service-listen.js";
 
 const sshLogger = logger;
 
@@ -47,14 +48,20 @@ registerDockerContainerRoutes(app, {
 
 const PORT = 30007;
 
-app.listen(PORT, "127.0.0.1", async () => {
-  try {
-    await authManager.initialize();
-  } catch (err) {
-    sshLogger.error("Failed to initialize Docker backend", err, {
-      operation: "startup",
-    });
-  }
+listenOnServicePort({
+  app,
+  port: PORT,
+  logger: sshLogger,
+  serviceName: "docker",
+  onListening: async () => {
+    try {
+      await authManager.initialize();
+    } catch (err) {
+      sshLogger.error("Failed to initialize Docker backend", err, {
+        operation: "startup",
+      });
+    }
+  },
 });
 
 process.on("SIGINT", () => {
--- a/src/backend/hosts/file-manager/index.ts
+++ b/src/backend/hosts/file-manager/index.ts
@@ -68,6 +68,7 @@ import { registerFileDownloadRoutes } from "./download-routes.js";
 import { registerFileActionRoutes } from "./action-routes.js";
 import { applyAgentAuth } from "../terminal-auth-helpers.js";
 import { applyCACertIfPresent } from "./ca-cert-auth.js";
+import { listenOnServicePort } from "../../utils/service-listen.js";
 
 /**
  * The host id came from whichever database the client is displaying. If this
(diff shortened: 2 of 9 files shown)


ZacharyZcR (approved):
Reviewed at c6db872e13a29c82e77dc1e0a958ba27bcdd93c1. The helper avoids Express 5's callback/error ambiguity, preserves the tunnel's existing server, and consistently makes fixed-port conflicts fatal and diagnosable. Local verification passed: 5 focused socket tests and full TypeScript project check.

PR #1389, opened 2026-09-04
feat(desktop): allow RDP/VNC/Telnet to originate from this device

nasif-naseef:
# Overview

RDP/VNC/Telnet were pinned to the remote server on the desktop, so a host the user's own machine can reach but the Termix server cannot was impossible to open — the app answered "Remote server required" even when no remote server was wanted.

- [x] Added: per-host **This device** origin for RDP, VNC and Telnet
- [x] Fixed: three call paths that would have silently ignored the setting
- [x] Updated: the Connection Origin control now appears for hosts that enable only these protocols

# Changes Made

Most of the local path already existed: the embedded desktop backend runs the Guacamole websocket server (`guac_init`, port 30008) and guacd's address is already configurable via `GUACD_URL`/`GUACD_HOST`/`GUACD_PORT`, the global guacd URL setting, and the per-host **guacd Proxy** override. What was missing was the choice.

`resolveConnectionOrigin` now honours an explicit per-host origin for these protocols. **Left on Default they still resolve to remote.** They need a guacd, which the desktop does not ship, so originating locally only works once the user has pointed Termix at one of their own — keeping it opt-in means an upgrade never moves a working connection onto a guacd that isn't there. This is deliberately asymmetric with SSH, where Default follows the desktop-wide setting; flagging it as the one product call in here, and happy to switch to full consistency if you'd rather. `serial` and non-Electron behaviour are unchanged.

Three paths had the same assumption baked in and would have quietly ignored the setting:

- The Guacamole call sites resolved the origin but never passed the host's own override, so it could never take effect.
- `guacamole-api` sent every token, `connect-host` and status call to the remote server whenever running under Electron, and remapped the host id onto the remote server's id — which would address the wrong row, or fail outright with no server configured.
- `GuacamoleDisplay` minted its token *before* resolving the origin, so the token could come from a different backend than the socket then dialled.

**The origin is a required parameter on those API functions rather than a defaulted one.** A default silently sent a missed call site to the remote server: the guacd status check in `fetchToken` was exactly that, and it failed with a bare "Network Error" on a desktop with no server configured. Making it required means the compiler names every caller instead — which is how `CollabRoomTab` was found; it now resolves from the host it already holds, so a locally-originated host stays local when presented into a room. It is the second parameter on `getGuacamoleTokenFromHost` because a required parameter cannot follow the optional ones, and it is as fundamental to the call as the host id.

Finally, the Connection Origin control was gated on `protocols.enableSsh`, so a host enabling only these protocols could never reach the setting. That gate is now `connectionOriginAppliesTo()`.

No new error handling was needed for an unreachable local guacd: `GuacamoleApp` has called `getGuacdStatus()` before every connection since v2.3.0 and reports `guacdUnavailable`. Making that call origin-aware was the whole fix — it now asks the backend the session will actually use, so the case is caught before a socket is opened. The error paths in `GuacamoleDisplay` are byte-identical to `dev-2.8.0`.

# Related Issues

- Fixes Termix-SSH/Support#1240 — cross-repo, so GitHub's closing keywords do not apply; the issue needs closing by hand on merge, as Support#1254 was for #1382.

# Screenshots / Demos

Verified end to end on Linux with a `guacamole/guacd:1.6.0` container on `127.0.0.1:4822`, against two real hosts set to **This device**.

**Connection reaches guacd and the target answers for itself:**

```
guacd[16]: User "@2dcab640…" joined connection "$87c9437b…" (1 users now present)
guacd[16]: Loading keymap "en-us-qwerty"
guacd[16]: RDP server closed/refused connection: ...
```

In the app, the two hosts report what their RDP servers actually said — `Server refused connection (wrong security type?)` and `Authentication failure (invalid credentials?)`. Both previously failed with "Remote server required" before anything was attempted.

**With guacd stopped**, the pre-existing status check reports it before connecting:

> Remote desktop service (guacd) is not available. Please ensure guacd is running and accessible and configured properly in admin settings.

One thing worth knowing for anyone trying this: a desktop that has ever been pointed at a Compose stack inherits `guac_url = guacd:4822`, which does not resolve outside Docker and fails with `getaddrinfo EAI_AGAIN guacd`. The guacd URL has to name something the desktop itself can reach. The field description now says so.

Tests: 12 new across `connection-origin` (Default→remote, explicit local, explicit remote, serial and web unchanged), `guacamole-api` (local origin uses the embedded backend and does not remap the host id; explicit remote unchanged; older behaviour preserved) and `HostEditorData` (the control applies to each protocol, and not when none is enabled).

`npx vitest run` goes from 2708 to 2719 passing with no new failures — the 119 pre-existing failures on `dev-2.8.0` are unrelated and I confirmed that baseline by stashing. `tsc -b`, `npm run lint` (0 errors, warning count unchanged at 102) and `prettier --check` are clean.

# Checklist


ZacharyZcR (approved):
Reviewed the local/remote Guacamole API selection, host-id mapping, token/WebSocket origin consistency, collaboration live-session path, editor exposure, and regression coverage at af85ca072ae006476fabd7196cc0ce91f787fc36. Focused tests pass (64/64) and the full TypeScript build passes.

PR #1391, opened 2026-09-04
feat(local-terminal): add copy/paste support

vietanhtwdk:
Local terminal had no clipboard wiring, so selected text couldn't be copied. Extracted the SSH terminal's copy/paste shortcut and right-click handling into a shared terminal-clipboard module and wired it into both terminals, removing duplicated logic in the process.

# Overview

Adds copy/paste support to the local terminal, which had no clipboard wiring at all. Also extracts the SSH terminal's copy/paste keyboard shortcut and right-click handling into a shared module so both terminals now share one implementation instead of duplicating the logic.

- [x] Added: clipboard copy/paste support for the local terminal (`Ctrl/Cmd+C`, `Ctrl+Insert`, `Ctrl+Shift+C`, `Ctrl+Shift+V`, and right-click copy/paste)
- [x] Updated: SSH terminal (`Terminal.tsx`) refactored to consume the same shared clipboard module instead of its own inline copy/paste logic
- [ ] Removed: ...
- [x] Fixed: selected text in the local terminal could not be copied

# Changes Made

- Added `src/ui/features/terminal/terminal-clipboard.ts` exporting:
  - `handleTerminalClipboardKeyEvent` — handles `Ctrl/Cmd+C` (copies selection, clears it), `Ctrl+Insert` (copies without clearing), `Ctrl+Shift+C`/`Ctrl+Shift+V` (always copy/paste), and a `plainPasteMode` option (`"native"` lets the browser's own paste event fire, used by the SSH terminal to avoid repeated OS clipboard permission prompts; `"explicit"` reads the clipboard directly, used by the local terminal, which has no native paste listener of its own)
  - `createTerminalContextMenuHandler` — right-click copy/paste, gated by the existing `rightClickCopyPaste` cookie preference, with an optional Ctrl+right-click passthrough (used by the SSH terminal to open the file manager)
- `LocalTerminal.tsx` now loads `ClipboardAddon`/`RobustClipboardProvider` and wires the shared key/context-menu handlers into its xterm instance
- `Terminal.tsx` (SSH) replaced its four inline copy/paste `if` blocks and its local context-menu handler with calls into the shared module — net removal of ~108 lines with no behavior change
- **Review fix:** the local terminal's session effect (which starts the PTY and whose cleanup closes it) depended on the `t` translation function. Since react-i18next gives `t` a new identity on every language change, switching languages was tearing down the running shell and spawning a new session. Translations are now read through a ref so localization is decoupled from the PTY lifecycle.
- Added `src/ui/tests/features/terminal/terminal-clipboard.test.ts` (16 cases) covering copy with/without selection, `Ctrl+Insert`, `Ctrl+Shift+C`/`V`, explicit vs. native plain `Ctrl+V`, the right-click preference cookie, and Ctrl+right-click passthrough

# Related Issues

- N/A

# Screenshots / Demos

- N/A

# Checklist


ZacharyZcR (changes requested):
The local-terminal clipboard work is useful, but this head introduces a session-lifecycle regression that needs fixing before merge.

`LocalTerminal` adds `t` to the effect that opens the local shell and whose cleanup calls `closeLocalTerminal(sessionId)`. react-i18next updates the translation function when the language changes, so changing the UI language will now tear down the running local shell and create a new session. Clipboard/toast localization must not be coupled to the PTY lifecycle; use a stable callback/ref or otherwise keep translation changes out of that session effect.

Please also add focused tests for the newly extracted `terminal-clipboard.ts` behavior (copy with/without selection, explicit vs native paste, right-click preference, and Ctrl+right-click passthrough). This refactors the existing SSH input path and adds a new local-terminal path, but the PR currently has no regression coverage for either.

ZacharyZcR (approved):
Re-reviewed the updated head after the requested changes. The translation function is now accessed through a ref and no longer participates in the PTY session effect lifecycle. The new clipboard regression suite passes 16/16; focused ESLint and the full TypeScript build also pass.

PR #1392, opened 2026-09-05
feat(desktop): local filesystem and transfer bridge for the file manager

maxser0v:
> **Review round 2 — restructured.** This PR is now *only* the Electron main-process bridge, so the trust boundary can be reviewed on its own (~850 lines incl. tests). The UI moved to follow-up PRs: #1413 (dual pane + drag-and-drop) → #1414 (local context menu) → #1393 → #1394 → #1395 → #1415 → #1396. Each of those is "previous branch + one commit".
>
> Review fixes are kept as separate commits so each delta is visible: `fix(desktop): harden the local transfer bridge` (c010285) for the two original blockers, and `fix(desktop): Windows-safe replace for overwrite downloads` (4a6bc67) for the rename-over-existing issue in the explicit overwrite path.

# Overview

Main-process side of a Termius-style Local | Remote file manager for the desktop app. No UI in this PR.

- [x] Added: `electron/local-files.cjs` — IPC handlers for browsing the local disk (home, list, mkdir, createFile, rename, trash, ensureDir, exists, walk, reveal, open) and for streaming files between the local disk and the file-manager backend (`uploadLocalFile`, `downloadToLocal`, progress events, cancel).
- [x] Added: `src/backend/tests/electron/local-files.test.ts` — 21 main-process tests.
- [x] Updated: `electron/preload.js` exposes `window.electronAPI.localFs` / `localTransfer`; the existing `invoke` allowlist is untouched.
- [x] Updated: `src/types/electron.d.ts`, `electron/main.cjs` (registration).

# Changes Made

**Trust boundary.** The renderer never supplies a URL or headers. It sends `{ origin: "local" | "remote", route, deviceId }` and the main process resolves the target itself:

- `route` must be one of a fixed allowlist (`uploadFileStream`, `downloadFileStream`).
- `origin: "local"` → the embedded backend base (`http://localhost:30004/ssh/file_manager`).
- `origin: "remote"` → the remote-sync config's URL (http/https only) with the stored JWT; refused if remote sync is not configured.
- `deviceId` is validated (`^[A-Za-z0-9._:-]{1,128}$`).
- Anything else throws before a request is made.

**Collision policy for downloads.** Nothing is renamed or replaced silently:

- Destination is checked first; `EEXIST` is returned unless the caller passes `overwrite: true`.
- Each transfer writes to its own `<dest>.<transferId>.termix-part`, opened with `wx`, and is published with `fs.link` / `COPYFILE_EXCL`. Two concurrent transfers to the same destination cannot share or clobber a partial — the second gets `EBUSY`.
- Partials are removed on failure and on cancel.

**Windows-safe overwrite.** `rename()` onto an existing name is POSIX-only (Windows fails with EEXIST/EPERM), so the explicit overwrite path never does it. The existing file is moved aside to a transfer-unique `<dest>.<transferId>.termix-replaced`, the partial is published exclusively under the freed name, then the aside copy is removed. If the aside step fails (file open on Windows) nothing is touched and the caller gets `EBUSY`; if publishing fails the original is moved back. Replacing a folder is refused with `EISDIR`. The publish primitives take an injectable `publishFs`, which is how the tests exercise this against a Windows-like `rename()` that refuses to overwrite.

**Streaming.** Uses Electron's `net` with `useSessionCookies` and `chunkedEncoding`, so the session cookie / remembered JWT is attached the same way as the renderer's own requests and large files are not buffered in memory.

**Testability.** `createLocalFileHandlers` / `createTargetResolver` take `net`, `shell` and the remote-sync getters as parameters, so the boundary is tested without Electron.

# Related Issues

- Followed by #1413 (UI), #1414, #1393, #1394, #1395, #1415, #1396 (stacked, each one commit).

# Screenshots / Demos

No UI. Test run:

```
npx vitest run src/backend/tests/electron/local-files.test.ts
 ✓ target resolution: local/remote/unknown origin, route allowlist, deviceId validation, unconfigured remote
 ✓ download: EEXIST without overwrite, overwrite replaces, EBUSY while in flight, unique partial per transfer, partial cleanup on failure/cancel, exists()
 ✓ upload: multipart body integrity (parsed with Busboy), off-origin refusal
 ✓ replace primitive (Windows-like fs): swap with no rename onto an occupied name, original restored on publish failure, EBUSY when the file cannot be moved aside, EISDIR for folders, vanished destination, end-to-end overwrite through the handler
 Tests  21 passed (21)
```

# Checklist

- [x] Code follows project style guidelines (prettier, eslint, type-check, build all green)
- [x] Supports mobile and desktop UI/app (desktop-only feature; no change for web/mobile)

### DIFF
--- a/electron/local-files.cjs
+++ b/electron/local-files.cjs
@@ -1,0 +1,632 @@
+// Local filesystem access + streamed local<->remote transfers for the
+// desktop app's dual-pane file manager.
+//
+// The renderer has no Node access (contextIsolation), so browsing the user's
+// own disk and moving bytes between it and the backend has to happen here.
+// Uploads/downloads are streamed through Electron's `net` stack against the
+// same file-manager HTTP routes the renderer already uses, so nothing is ever
+// buffered whole in memory and the existing auth (session cookies + bearer
+// JWT supplied by the renderer) keeps working unchanged.
+
+const fs = require("fs");
+const fsp = require("fs/promises");
+const os = require("os");
+const path = require("path");
+const { URL } = require("url");
+const { net } = require("electron");
+
+const IPC = {
+  HOME: "local-fs:home",
+  LIST: "local-fs:list",
+  MKDIR: "local-fs:mkdir",
+  CREATE_FILE: "local-fs:create-file",
+  RENAME: "local-fs:rename",
+  TRASH: "local-fs:trash",
+  ENSURE_DIR: "local-fs:ensure-dir",
+  WALK: "local-fs:walk",
+  REVEAL: "local-fs:reveal",
+  OPEN: "local-fs:open",
+  UPLOAD: "local-transfer:upload",
+  DOWNLOAD: "local-transfer:download",
+  CANCEL: "local-transfer:cancel",
+  PROGRESS: "local-transfer:progress",
+};
+
+const READ_CHUNK_BYTES = 1024 * 1024;
+const PROGRESS_INTERVAL_MS = 150;
+
+const activeTransfers = new Map();
+
+function isAbsoluteLocalPath(candidate) {
+  return typeof candidate === "string" && path.isAbsolute(candidate);
+}
+
+function normalizeLocalPath(candidate) {
+  if (!isAbsoluteLocalPath(candidate)) {
+    throw new Error("A local absolute path is required");
+  }
+  return path.normalize(candidate);
+}
+
+function isHiddenEntry(name) {
+  return name.startsWith(".");
+}
+
+// A single path segment: no separators, not "." / "..", no NULs.
+function requireEntryName(name, what) {
+  const safeName = String(name || "").trim();
+  if (
+    !safeName ||
+    safeName === "." ||
+    safeName === ".." ||
+    /[\/\\\0]/.test(safeName)
+  ) {
+    throw new Error(`Invalid ${what}`);
+  }
+  return safeName;
+}
+
+async function pathExists(target) {
+  try {
+    await fsp.lstat(target);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+async function describeEntry(dirPath, dirent) {
+  const entryPath = path.join(dirPath, dirent.name);
+  let type = "file";
+  let size = 0;
+  let modifiedTimestamp;
+  let linkTarget;
+
+  try {
+    if (dirent.isSymbolicLink()) {
+      type = "link";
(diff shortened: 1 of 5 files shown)


ZacharyZcR (changes requested):
There are two blocking correctness/security issues in the new main-process transfer boundary:

1. `downloadToLocal()` always writes to `${dest}.termix-part` and then calls `rename(partialPath, dest)`. On POSIX that silently replaces an existing local file, while Windows commonly fails when the destination exists. Two concurrent downloads to the same destination also share the same partial file and can corrupt each other. Please define an explicit collision policy (prefer refusing/asking rather than destructive replacement) and use a transfer-unique temporary sibling, with regression tests.
2. The privileged Electron handler accepts an arbitrary renderer-supplied URL and arbitrary headers, then performs the request with session cookies. That turns the main process into an unrestricted HTTP/SSRF primitive and permits sending Termix auth headers to an arbitrary origin. Please derive or strictly validate the target against the configured local/remote Termix origin in the main process; do not trust the renderer-provided URL/header set.

The focused tests currently pass (24 tests), but neither boundary is covered. Please add main-process tests for destination collisions/concurrency and rejected off-origin targets before this can merge.

ZacharyZcR (changes requested):
The two original blockers are addressed well on this head: renderer-controlled targets/headers are no longer trusted, and non-overwrite downloads now use exclusive, transfer-specific publication with focused race tests. One cross-platform blocker remains in the new explicit overwrite path: `publishDownload()` uses `fsp.rename(partialPath, absDest)` when `overwrite` is true. POSIX replaces the destination, but Windows rename commonly fails with `EEXIST`/`EPERM` when the destination exists, so the Replace action does not actually work on a primary desktop platform. Please implement a Windows-safe replacement strategy that preserves the original on failure as far as practical, and add a focused test that exercises the replace primitive without relying on POSIX rename-over-existing semantics. The dependent stack should remain blocked until this root PR is corrected.

ZacharyZcR (approved):
The Windows overwrite blocker is resolved on this exact head. The replacement no longer renames onto an occupied destination, preserves the original when publication fails, rejects directories/locked files explicitly, and carries focused Windows-like plus end-to-end coverage. I independently ran the focused suite with loopback permission (21/21), TypeScript type-check, and Prettier checks successfully. Approved as the root of the stack; dependent PRs still need rebasing onto the merged root and separate incremental review.

PR #1395, opened 2026-09-05
feat(file-manager): show/hide list-view columns from the header

maxser0v:
> **Stacked on #1394.** Only `feat(file-manager): show/hide list-view columns from the header` (465b88a) is new here — please review that commit alone. Will rebase onto `dev-2.8.0` once #1394 merges.

# Overview

Show/hide list-view columns from the header.

- [x] Added: right-click a list-view header (remote grid or local pane) to open a Columns menu and tick/untick Modified, Owner, Size, Permissions (remote) or Modified, Size, Kind (local).
- [x] Added: hidden columns leave the grid template entirely so the Name column gets the space back. At least one optional column always stays on.
- [x] Added: the choice is remembered per pane alongside the column widths.

# Changes Made

- New `ColumnVisibilityMenu` component; `useResizableColumns` extended with visibility state and tests.
- Two new i18n keys in `en.json`.

Rebased on current `dev-2.8.0`.

# Related Issues

- Depends on #1394 (→ #1393 → #1414 → #1413 → #1392). Followed by #1415.

# Screenshots / Demos

image

# Checklist


ZacharyZcR (dismissed):
This is described as stacked on #1394, but it targets , repeats the full lower stack, contains unresolved  references, and is currently conflicting. Please retarget it to  (or reduce the diff to only the column-visibility commit) after the lower stack is fixed. I will review the isolated delta then.

ZacharyZcR (changes requested):
This is described as stacked on #1394, but it targets `dev-2.8.0`, repeats the full lower stack, contains unresolved `#TODO` references, and is currently conflicting. Please retarget it to `feature/file-manager-parent-entry` (or reduce the diff to only the column-visibility commit) after the lower stack is fixed. I will review the isolated delta then.

ZacharyZcR (changes requested):
Request changes: remote filenames can escape the selected download directory on Windows.

`useLocalTransfers.downloadRemoteItems` builds local destinations by splitting relative remote paths on `/` and passing each remote name into `joinLocalPath`. That helper concatenates the name without rejecting Windows path separators or traversal. The Electron download handler then calls `path.normalize(destPath)` and creates/writes the normalized destination; it has no selected-root containment check.

Reproduced with the actual helper and Windows path normalization: selected directory `C:\Downloads\selected`, remote POSIX filename `..\outside.txt` -> local destination `C:\Downloads\outside.txt`. Backslash is a valid character in a POSIX filename. A newly downloaded file can therefore be created outside the user's selected directory, even when overwrite is disabled.

Please validate remote path components for the destination platform and enforce containment under the selected local root before filesystem operations. Add Windows regression tests for backslash traversal, absolute/drive-qualified names and ordinary nested folders.

This implementation is shared unchanged by the PR chain #1413 -> #1414 -> #1393 -> #1394 -> #1395 -> #1415 -> #1396. Fix the base transfer PR, propagate the fix through dependent branches, and rerun checks before merging any of these heads. The unrelated column/sidebar features are not the cause, but their current heads include the same vulnerable download code.

ZacharyZcR (changes requested):
Request changes after re-review of the updated head: lexical traversal is fixed, but filesystem containment still fails through existing symlinks.

The new `assertWithinRoot` uses resolve/normalize/relative on path strings. It does not inspect the filesystem. Both download publication and the guarded ENSURE_DIR handler subsequently follow existing parent-directory symlinks.

Reproduced by invoking the actual updated ENSURE_DIR IPC handler against disposable directories: `selected/linked` is an existing symlink to a sibling `outside` directory; requesting `selected/linked/remote-subdir` with rootPath=`selected` returns success and creates `outside/remote-subdir`. All remote path components in this example are ordinary names, so renderer component validation does not prevent it. The download destination check uses the same lexical guard, followed by filesystem operations that resolve the symlink.

Please enforce containment using the selected root's resolved filesystem location and the destination's existing parent chain, including symlink/junction cases, and avoid a check/write race. Add regression tests exercising the actual directory and download handlers through an existing link, while preserving explicitly selected linked roots if supported.

The original Windows `..\outside.txt` case is fixed and the focused existing tests pass (44 tests). This remaining bypass still prevents safe merge. All seven dependent PR heads currently share the same `electron/local-files.cjs` implementation; please propagate the correction through the chain.

ZacharyZcR:
Maintainer repair pushed at `9c26824d052f7291cc8c19d967b755d305fa452e`. Original contributor history is preserved; current `dev-2.8.0` is integrated.

Integrated the file-manager download containment repair from #1413. Descendant symlinks/junctions are rejected, temporary downloads stay in the selected root, and destination containment is checked again before publication. Tests cover an existing link and a parent replaced during streaming, while permitting an explicitly selected linked root.

Validation: 100 focused tests passed. `npm run type-check`, `npm run lint` (0 errors; existing warnings), and `npm run build` passed locally. CI is being approved for the new head. This update does not mark the PR approved or merged.

ZacharyZcR (approved):
Re-reviewed repaired head 9c26824d052f7291cc8c19d967b755d305fa452e. The previously reported blockers are resolved. Both PR Check jobs pass on this exact head. The combined ten-PR integration tree passed 215 focused tests, type-check, lint, Prettier and production build. Approving the scoped change after checking its dependency delta and current dev merge.

PR #1399, opened 2026-09-06
fix(terminal): persist Ctrl+/- font zoom so it survives option refreshes

tcezarl:
# Persist terminal font zoom (Ctrl +/- / Ctrl+wheel) so it survives option refreshes

## Problem

Zooming the terminal with `Ctrl` `+`/`-` or `Ctrl`+wheel changes the font size for
~30–60 s and then snaps back to the configured size.

`changeTerminalFontSize()` only mutates `terminal.options.fontSize` and
`terminalFontSizeRef.current`; it never persists. The "update terminal options
individually" effect in `Terminal.tsx` (deps include `isFitted`,
`terminalDefaults`, `hostConfig.terminalConfig`) re-runs on the WebSocket
keepalive / refit / reconnect cycle and unconditionally reassigns
`terminal.options.fontSize = config.fontSize`, discarding the zoom.

## Fix

Persist the zoom level per host in `localStorage`
(`terminal_fontsize_host_<hostId>`, mirroring the existing
`terminal_theme_host_<hostId>` pattern) and have the terminal-option effects
prefer that override:

* `changeTerminalFontSize()` writes the new size to `localStorage`.
* The creation effect and the "update options" effect read the override and
  apply it instead of `config.fontSize`.
* If the configured font size itself changes (user edits it in Settings /
  host config), the override is cleared so the new configured value wins.

All storage access is wrapped in `try/catch` (private mode / disabled storage
falls back to the configured size).

## Testing

* Zoom in, wait past the keepalive interval — size holds.
* Reload — size persists.
* Change Font Size in Connection Defaults — override is dropped, new value applies.
* Storage disabled — no errors, behaves as before (transient zoom only).


ZacharyZcR (changes requested):
This needs changes before merge. The persisted value stores only the zoomed font size, not the configured size it was derived from. If a user zooms a host, closes the terminal, changes Connection Defaults while that terminal is unmounted, and reopens it, `lastConfiguredFontSizeRef` initializes to the new configured value and the old localStorage override still wins forever. The claimed “configured font size change clears the override” only works while this component remains mounted. Store the configured/base size alongside the override (or use another design that can invalidate stale overrides on mount) and add a regression test for that closed-terminal settings-change path.

Also avoid `terminal_fontsize_host_undefined`: quick-connect terminals do not necessarily have a host id, so unrelated quick-connect sessions currently share one persisted override. Finally, please retarget/rebase this onto the active `dev-2.8.0` branch; the current `main` base produces an unrelated multi-file diff against development.

ZacharyZcR (changes requested):
Request changes: the persisted font zoom deletes itself on remount.

In `Terminal.tsx`, initialization captures `configuredFontSize`, then overwrites `config.fontSize` with the saved zoom. The options-update effect compares that already-overridden `config.fontSize` with `lastConfiguredFontSizeRef.current`, initialized from the original configured size, and removes the localStorage value when they differ.

Reproduced using the actual initialization/update blocks: configured 14px + saved 18px -> first mount displays 18px but deletes the saved value -> next mount falls back to 14px.

Keep the configured base size separate from the effective zoom size, and compare the configured value when deciding whether defaults changed. Add regression coverage for zoom -> close -> reopen -> close -> reopen, and for explicitly changing the configured font size. The previous base/conflict issues are resolved; this functional defect remains at the current head.

ZacharyZcR:
Maintainer repair pushed at `bfa63b3cebd0673226f1e556ca62ce115ace9ccf`. Original contributor history is preserved; current `dev-2.8.0` is integrated.

Persisted zoom now records the configured font size together with the override. Reopening a terminal preserves zoom; changing the configured size invalidates stale zoom, including while unmounted. Missing host identities do not share a storage key, and live handlers use the current configuration.

Validation: 98 focused tests passed. `npm run type-check`, `npm run lint` (0 errors; existing warnings), and `npm run build` passed locally. CI is being approved for the new head. This update does not mark the PR approved or merged.

ZacharyZcR (approved):
Re-reviewed repaired head bfa63b3cebd0673226f1e556ca62ce115ace9ccf. The previously reported blockers are resolved. Both PR Check jobs pass on this exact head. The combined ten-PR integration tree passed 215 focused tests, type-check, lint, Prettier and production build. Approving the scoped change after checking its dependency delta and current dev merge.

PR #1400, opened 2026-09-06
docs: point security note to the docs root (old /security page is gone)

ssmurfgg04-gif:
## What
The security paragraph links to docs.termix.site/security, a page that no longer exists. I pointed the link at the docs root in the main README and in every translated README.

## Why
The /security path returns 404. The docs site has no security page in its nav, so the root is the correct target for now.

## Verification
docs.termix.site/security returns 404, docs.termix.site returns 200. Same sentence verified in each translated file.


ZacharyZcR (changes requested):
The dead `/security` URL does need fixing, but the replacement currently redirects to `/install/`, so it no longer supports the README sentence that says readers can learn how per-user secret and database encryption work. Please point these links to the existing security page instead: `https://docs.termix.site/features/authentication/security/`. I verified that page returns 200 and contains the relevant Database Encryption and secret-encryption documentation. Please also squash the ten identical one-file commits before merge (or let the repository squash-merge them); no need to change the content separately per locale beyond the URL.

ssmurfgg04-gif:
Thanks for the thorough review, the redirect to /install/ dropping the encryption context is exactly the problem. I rebased on current main first, and in the meantime the same dead /security link had spread to the Russian, Korean, Turkish and Vietnamese readmes, so the PR now covers all 14 files, English plus 13 translations, in a single commit. Every link now points at the authentication security page, I went with the trailing-slash URL because the bare one 308 redirects to it, and I confirmed the page returns 200 and actually documents the database and secret encryption. Ready for another look whenever you get a chance.

ZacharyZcR (approved):
The requested correction is now present on this exact head: every README locale points to the existing security documentation, and the target returns HTTP 200. The remaining repeated commits are harmless under squash merge. Approved.

PR #1403, opened 2026-09-06
fix: use i18n for host status tooltip labels

slsgzs-cloud:
## Summary

Fixes #1265.

### The bug

`buildStatusTooltip` has hardcoded English strings that bypass i18n:

```
const statusLabel =
  status === "online" ? "Available"
  : status === "reachable" ? "Reachable, not authenticated"
  : "Offline";
if (!statusCheckEnabled(host)) return "Monitoring disabled";
```

When the user's language is set to Russian (or any non-English language), these tooltips still render in English, while every other string on the same row is translated.

### Fix

1. Added `t` parameter to `buildStatusTooltip` with a default that returns the key (for backward compatibility in tests)
2. Replaced hardcoded strings with `t()` calls using new keys:
   - `hosts.status.available`
   - `hosts.status.reachable`
   - `hosts.status.offline`
   - `hosts.status.monitoringDisabled`
3. Updated the call site to pass `t` from `useTranslation()`

### New translation keys

| key | default English value |
|---|---|
| `hosts.status.available` | Available |
| `hosts.status.reachable` | Reachable, not authenticated |
| `hosts.status.offline` | Offline |
| `hosts.status.monitoringDisabled` | Monitoring disabled |

These should be added to the translation files. Until then, the fallback to the key name means English-speaking users see the key path instead of the English text — but every other language sees translated text.


ZacharyZcR (changes requested):
This does not yet fix #1265 safely. The code starts calling four new `hosts.status.*` keys, but this PR adds none of those keys to `src/ui/locales/en.json` (or the source translation catalogue). The default translator argument only affects direct helper calls in tests; the real `t()` call will render the key path when a resource/fallback is missing, exactly as the PR body notes. Please add the canonical English entries, update focused tests to assert translated labels rather than key paths, and retarget the active development branch (`dev-2.8.0`) unless this is intentionally a stable-only patch.

ZacharyZcR (changes requested):
The canonical English keys are now present, so the original missing-resource defect is fixed. Two requested items still remain on this head: retarget the active development branch (`dev-2.8.0`) and add/update focused tests that exercise `buildStatusTooltip` with a translator and assert the rendered labels rather than key paths. Please also restore the final newline in `en.json`.

ZacharyZcR (changes requested):
The requested focused tests are now present and the PR targets `dev-2.8.0`, so those review items are addressed. This exact head is currently `CONFLICTING/DIRTY` against the active branch and cannot be validated or merged as-is. Please rebase it onto the current `dev-2.8.0`, resolve the locale/HostItem conflicts without dropping the canonical English keys or tests, and restore the final newline in `en.json` if it is still missing. I will re-review the resulting head.

PR #1404, opened 2026-09-07
Surface SSH host key changes in host status

ZacharyZcR:
## Summary
When a stored SSH host key changes, metrics polling pauses and the host tooltip explains that verification is required. After the user accepts the replacement key and it is successfully saved, polling can resume.

- Track host-key verification failures separately from ordinary authentication/network failures and preserve the reason across TCP reachability checks.
- Clear only the affected host's HOST_KEY failure after saving a verified replacement. Rejected keys, failed saves, and unrelated AUTH/TOTP failures remain blocked.
- Preserve the translated status tooltip when integrating current `dev-2.8.0`.

## Validation
- 46 focused host-key, polling state, host status, status-store, and tooltip tests passed.
- Regression tests cover accepting/rejecting replacement keys, database write failure, and preserving AUTH/TOTP failures.
- `npm run type-check`, `npm run lint`, `npm run build` passed.

Fixes Termix-SSH/Support#1266

### DIFF
--- a/src/backend/hosts/metrics/host-status.test.ts
+++ b/src/backend/hosts/metrics/host-status.test.ts
@@ -1,8 +1,24 @@
 import { describe, expect, it } from "vitest";
 import {
+  isHostKeyVerificationError,
   statusAfterAuthentication,
   statusAfterReachabilityCheck,
 } from "./host-status.js";
+
+describe("isHostKeyVerificationError", () => {
+  it.each([
+    "Host denied (verification failed)",
+    "Host key changed - please connect via Terminal to verify the new key",
+  ])("classifies %s", (message) => {
+    expect(isHostKeyVerificationError(new Error(message))).toBe(true);
+  });
+
+  it("does not classify ordinary authentication failures", () => {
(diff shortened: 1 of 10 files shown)


ZacharyZcR:
Maintainer audit: changes required before merge (author-account limitation prevents a formal Request Changes review).

The new HOST_KEY failure is recorded as permanent. `shouldSkip` then prevents future metric authentication and `cleanup` retains the permanent record. The only production call to `authFailureTracker.reset` is after successful metric collection, which cannot be reached once polling is skipped. Accepting the replacement host key, refreshHostPolling, and refreshAllPolling do not reset this state.

Result: after a host-key mismatch, the user can accept the new key in Terminal but metrics remain suspended for the lifetime of the process.

Please reset the relevant host failure state after verified key acceptance or an explicit authorized retry, with a regression test covering mismatch -> accept new key -> metrics resume. Also resolve the current conflict with dev's translated status tooltip, preserving its translator argument. Keep this PR unmerged pending the recovery fix and fresh checks.

PR #1413, opened 2026-09-08
feat(file-manager): Termius-style Local | Remote dual pane with drag-and-drop transfers (desktop)

maxser0v:
> **#1392 (main-process bridge) is merged**; this branch is rebased onto `dev-2.8.0`. Two commits: the feature (`feat(file-manager): Termius-style Local | Remote dual pane…`, 89bdac5) and the review fix `fix(file-manager): keep downloaded files inside the selected local folder` (6c1ed04), kept separate so the delta is visible. Ready for review.

# Overview

Renderer side of the dual-pane file manager: a Local pane next to the remote grid, with drag-and-drop transfers in both directions. Desktop app only — the toggle is hidden when `window.electronAPI.localFs` is absent, so the web build is unchanged.

- [x] Added: `LocalFilePane` (grid/list views, hidden files, breadcrumbs, New Folder); toolbar toggle (Laptop icon); remembered path/visibility/width.
- [x] Added: `useLocalTransfers` — batch upload/download through the main process with one progress toast per batch (speed, ETA, cancel).
- [x] Added: Replace / Skip prompt for downloads that would collide with existing local files.
- [x] Updated: `FileManagerGrid` accepts local-file drops (upload) and marks its internal drags so the Local pane can accept them (download).
- [x] Updated: `en.json` — new `fileManager.local*` keys (English only; other locales via Crowdin).

# Changes Made

- **Drag contract.** Local drags carry `application/x-termix-local-files` (JSON `{ type: "local_files", paths }`); the remote grid's internal drags carry an `application/x-termix-remote-files` marker. Each pane distinguishes local / remote / Finder drops by these, and Finder drops on the remote grid keep working exactly as before.
- **Transfers.** `src/ui/api/local-transfer-api.ts` describes a target as `{ origin, route, deviceId }` (`getSessionOrigin(sessionId)` in `main-axios.ts`) — the renderer never hands the main process a URL. Folders are walked in the main process and uploaded file-by-file with relative paths; empty folders are recreated.
- **Collision policy.** Before a download batch starts, destinations are checked (`localPathsExist`). If any exist the user gets a toast with **Replace** / **Skip**. Skip, dismiss and timeout all mean skip — nothing is replaced without an explicit click, and the main process enforces the same rule.
- **Destination containment.** Every remote path component is validated for the destination platform before it becomes a local path (`assertSafeLocalComponent`: no separators, traversal or NUL anywhere; on Windows also no `\ : * ? " < > |`, control chars, trailing dot/space or reserved device names) and the joined path is asserted to stay strictly under the selected folder (`buildLocalDestination`). Items that fail are skipped before any filesystem call and reported. Independently, the main process requires the selected folder (`rootPath`) and refuses any destination that resolves outside it (`assertWithinRoot`, `EINVAL`) before the request is sent or a file is created; the directory skeleton (`ensureDir`) gets the same check.
- **Persistence.** `termix:file-manager:local-pane:{path,hidden,open,width}` in localStorage.
- **Tests.** `LocalFilePane.test.tsx` (rendering, navigation, drop handling), `local-transfer-utils.test.ts` (relative-path planning, size formatting, destination safety: Windows backslash traversal, absolute/drive-qualified/UNC names, reserved names, ordinary nested folders), and `src/backend/tests/electron/local-files.test.ts` gains a containment group run against `path.win32` plus an end-to-end refused download with no network or disk side effects.

# Related Issues

- Builds on #1392 (merged). Followed by #1414 (local context menu) → #1393 → #1394 → #1395 → #1415 (sidebar toggle) → #1396 (stacked, one commit each).

# Screenshots / Demos

Collision prompt for a download batch that would overwrite existing local files (Skip / dismiss / timeout all mean skip):

replace-skip-prompt

How to test (desktop): `npm run dev` + `npm run electron:dev` → File Manager → connect a host → click the Laptop icon → drag a file from Local onto the remote grid (upload) and a remote file onto Local (download). Drop a remote file into a folder where it already exists to see Replace / Skip.

# Checklist

- [x] Supports mobile and desktop UI/app (desktop-only; hidden elsewhere)


ZacharyZcR (changes requested):
Request changes: remote filenames can escape the selected download directory on Windows.

`useLocalTransfers.downloadRemoteItems` builds local destinations by splitting relative remote paths on `/` and passing each remote name into `joinLocalPath`. That helper concatenates the name without rejecting Windows path separators or traversal. The Electron download handler then calls `path.normalize(destPath)` and creates/writes the normalized destination; it has no selected-root containment check.

Reproduced with the actual helper and Windows path normalization: selected directory `C:\Downloads\selected`, remote POSIX filename `..\outside.txt` -> local destination `C:\Downloads\outside.txt`. Backslash is a valid character in a POSIX filename. A newly downloaded file can therefore be created outside the user's selected directory, even when overwrite is disabled.

Please validate remote path components for the destination platform and enforce containment under the selected local root before filesystem operations. Add Windows regression tests for backslash traversal, absolute/drive-qualified names and ordinary nested folders.

This implementation is shared unchanged by the PR chain #1413 -> #1414 -> #1393 -> #1394 -> #1395 -> #1415 -> #1396. Fix the base transfer PR, propagate the fix through dependent branches, and rerun checks before merging any of these heads. The unrelated column/sidebar features are not the cause, but their current heads include the same vulnerable download code.

maxser0v:
Thanks — fixed in `fix(file-manager): keep downloaded files inside the selected local folder` (6c1ed04), kept as a separate commit on this PR.

Reproduced your case first (selected `C:\Downloads\selected`, remote name `..\outside.txt`): the old code built `C:\Downloads\selected\..\outside.txt`, which `path.win32.normalize` turns into `C:\Downloads\outside.txt`, with no containment check anywhere.

Two independent layers now enforce it:

- **Renderer** — `assertSafeLocalComponent(name, separator)` validates each remote path component for the destination platform (never empty / `.` / `..`, never `/` or NUL; on Windows additionally no `\ : * ? " < > |`, control chars, trailing dot/space, reserved device names), and `buildLocalDestination` asserts the joined path is strictly under the selected folder (case-insensitive on Windows). Files and directory entries that fail are skipped before any filesystem call and the user is told how many. On macOS/Linux a backslash stays a legal character, and the result is still inside the folder.
- **Main process** — `downloadToLocal` now requires `rootPath` (the selected folder) and `assertWithinRoot` refuses any destination that is the root itself, escapes it after normalisation, is absolute relative to it (other drive, UNC) or contains empty/`..` segments — `EINVAL`, before the request is made or a partial is created. `ensureDir` applies the same check when a root is passed (directory skeleton of downloaded trees). `path` is injectable so the Windows rules run in the tests on every OS.

Regression tests (renderer + `path.win32`): backslash traversal (`selected\..\outside.txt` → refused), absolute / drive-qualified / UNC names, ordinary nested folders accepted incl. case differences, reserved names / trailing dots / control chars, POSIX traversal, and an end-to-end handler test that a `../outside.txt` download and an escaping ensure-dir are refused with zero network traffic and nothing written. 107 file-manager/bridge tests, lint, type-check, prettier and build pass at the top of the chain.

#1414 → #1393 → #1394 → #1395 → #1415 → #1396 are re-stacked on this head (each still one commit of their own) and re-checked.

ZacharyZcR (changes requested):
Request changes after re-review of the updated head: lexical traversal is fixed, but filesystem containment still fails through existing symlinks.

The new `assertWithinRoot` uses resolve/normalize/relative on path strings. It does not inspect the filesystem. Both download publication and the guarded ENSURE_DIR handler subsequently follow existing parent-directory symlinks.

Reproduced by invoking the actual updated ENSURE_DIR IPC handler against disposable directories: `selected/linked` is an existing symlink to a sibling `outside` directory; requesting `selected/linked/remote-subdir` with rootPath=`selected` returns success and creates `outside/remote-subdir`. All remote path components in this example are ordinary names, so renderer component validation does not prevent it. The download destination check uses the same lexical guard, followed by filesystem operations that resolve the symlink.

Please enforce containment using the selected root's resolved filesystem location and the destination's existing parent chain, including symlink/junction cases, and avoid a check/write race. Add regression tests exercising the actual directory and download handlers through an existing link, while preserving explicitly selected linked roots if supported.

The original Windows `..\outside.txt` case is fixed and the focused existing tests pass (44 tests). This remaining bypass still prevents safe merge. All seven dependent PR heads currently share the same `electron/local-files.cjs` implementation; please propagate the correction through the chain.

ZacharyZcR:
Maintainer repair pushed at `fe60dc404df26886917864e97a42fa1419e690d3`. Original contributor history is preserved; current `dev-2.8.0` is integrated.

Integrated the file-manager download containment repair from #1413. Descendant symlinks/junctions are rejected, temporary downloads stay in the selected root, and destination containment is checked again before publication. Tests cover an existing link and a parent replaced during streaming, while permitting an explicitly selected linked root.

Validation: 88 focused tests passed. `npm run type-check`, `npm run lint` (0 errors; existing warnings), and `npm run build` passed locally. CI is being approved for the new head. This update does not mark the PR approved or merged.

ZacharyZcR (approved):
Re-reviewed repaired head fe60dc404df26886917864e97a42fa1419e690d3. The previously reported blockers are resolved. Both PR Check jobs pass on this exact head. The combined ten-PR integration tree passed 215 focused tests, type-check, lint, Prettier and production build. Approving the scoped change after checking its dependency delta and current dev merge.

PR #1416, opened 2026-09-08
feat(hosts): add per-host Web Endpoints (direct or SSH-tunnelled web UIs)

nasif-naseef:
## Summary

Adds per-host Web Endpoints for direct HTTP(S) access or an on-demand SSH tunnel, with separate browsing sessions for untrusted endpoint pages.

- Host editor, sidebar picker, backend validation and database fields configure endpoints and their access mode.
- Tunnel opening reports the actual bound port, enforces host access, and reopens changed targets; unused tunnels expire.
- Embedded tabs use credentialless frames with an opaque sandbox origin. Unsupported browsers fail closed before opening a tunnel.
- Desktop isolated windows use a unique in-memory Electron session without Termix preload or Node integration. Endpoint storage survives reloads within that window; login popups inherit the isolated session. Closing the endpoint closes its popups and clears session storage/connections.
- Non-web navigation, native protocols, permissions and automatic downloads are denied. Certificate exceptions are restricted to the explicitly selected origin, including its port, and cannot inherit Termix's global certificate exceptions.

Browser deployments retain embedded-mode limitations for sites requiring origin storage or popup login. Those pages can use the desktop isolated-window option. This feature does not open endpoint URLs in the user's shared browser session.

Related: Termix-SSH/Support#1273.

## Validation

- 182 focused endpoint, UI, routing, validation and isolation tests passed.
- Type-check, lint (0 errors; 102 existing warnings), Prettier and production build passed.
- Bounded real Electron smoke test: redirect requests carried no Termix cookie, endpoint Set-Cookie left the parent session unchanged, endpoint localStorage survived reload, and the endpoint had neither Node nor the Termix bridge.
- Popup inheritance, certificate origin boundaries, main-frame-only IPC authorization and cleanup have focused regression coverage.

Current development branch integrated. Final CI and maintainer review apply to the latest head.


ZacharyZcR (changes requested):
Request changes: the direct web endpoint path can leak the Termix session cookie to another service on the same host.

`unreachableTunnelReason` returns null for every `access: direct` endpoint. `WebEndpointTab` then embeds the URL generated from the host address and endpoint port. The cookie-isolation guard applies only to tunnel endpoints.

Reproduced using the actual URL/refusal helpers and a cookie jar with a dummy Secure/HttpOnly/SameSite=Lax `jwt` cookie: a Termix page at `https://termix.example/` allows a direct endpoint at `https://termix.example:8443/`, and the cookie is included for that destination. Cookie scope does not isolate ports. The target service receives the session token despite being a different origin.

Please apply session-cookie isolation to direct endpoints as well as tunnels, before navigation, including any supported redirects. Add regression coverage for same-host/different-port direct endpoints and retain the existing tunnel checks. Keep this PR unmerged until the credential boundary is addressed.

nasif-naseef:
Confirmed and fixed in e7f0ffc.

The finding is accurate: `unreachableTunnelReason` returned null for every `access: direct` endpoint, so the guard was tunnel-only and `WebEndpointTab` framed the host address unconditionally. Same leak as the tunnel case, reached without a tunnel.

**What changed**

- Renamed `unreachableTunnelReason` → `webEndpointRefusalReason`. The old name is a large part of why this gap existed — it read as tunnel-scoped, so the direct path never got the check.
- It now takes the endpoint's target host. A direct endpoint is refused in a browser when that host shares Termix's cookie site, evaluated **before** navigation in *both* open paths — the embedded tab and `openWebEndpointExternally`. Refusing after framing would already have leaked on the first request.
- New `sharesCookieSiteWithPage(targetHost, pageHost)`:
  - equal host → same-site. The web client sets `jwt` host-only (`path=/`, no `Domain`), so the port is irrelevant — this is your `https://termix.example/` vs `https://termix.example:8443/` case.
  - parent/sub domain → also same-site, covering the reverse direction where the target answers `Set-Cookie: jwt=...; Domain=<parent>` and lands in the jar both auth middlewares read before the `Authorization` header.
  - the suffix test is anchored on a dot boundary, so `eviltermix.example` does not pass as `termix.example`.
- Desktop stays exempt. I verified at runtime (CDP `Network.getAllCookies` against both a dev-mode and a packaged `file://` build) that the desktop session jar carries no `jwt` at all — auth there is Bearer-only from localStorage, so there is no cookie to leak.
- Dropped "or change this endpoint to Direct" from the tunnel refusal copy. It was pointing users at the path that had this exact bug.

**Coverage**

12 new tests. Regression cases for same-host/different-port direct endpoints, sub/parent-domain direct, the label-boundary bypass, and direct-on-desktop; all existing tunnel checks retained unchanged. I verified each new test fails against the pre-fix code (removing the direct branch fails exactly those 4 assertions and nothing else) so they are not passing for an unrelated reason.

Suite: 3092 passing, 1 skipped, 0 failing. `tsc -b --force` clean, lint 0 errors (102 warnings, unchanged), `prettier --check` clean.

**One residual, stated plainly**

Sibling subdomains under a shared parent (`a.example.com` served Termix, direct endpoint at `b.example.com`) are not caught. Detecting that needs a real registrable-domain (eTLD+1) test, which needs a public-suffix list this app does not currently bundle. I did not add that dependency on my own initiative — say the word if you would rather I pull in `tldts`/`psl` and make the check exact, and I will.

On redirects: the pre-navigation guard covers the URL we generate. Once the browser follows a cross-origin redirect inside the frame we no longer control the destination, so I did not claim coverage there.

ZacharyZcR (changes requested):
Request changes after re-review: the exact same-host case is fixed, but sibling-domain cookie injection remains possible.

`sharesCookieSiteWithPage` checks equality and parent/child hostname suffixes, so `panel.example.com` is allowed from a Termix page at `termix.example.com`. The endpoint can still return `Set-Cookie: jwt=...; Domain=example.com; Path=/users; Secure; SameSite=Lax`, which reaches Termix's /users requests. A narrower cookie path can place that cookie ahead of the existing host-only root-path session cookie.

Reproduced with the actual updated refusal helper and a cookie jar using dummy tokens: the sibling endpoint is allowed; the injected parent-domain cookie becomes the first jwt cookie for `https://termix.example.com/users/me`. Thus parent/subdomain checks do not fully establish the credential boundary described by this change. Initial-URL checks also do not constrain redirects from an otherwise allowed endpoint.

Please isolate the endpoint's browsing credentials or enforce a complete cookie-domain/navigation policy, including sibling subdomains and redirects. Add tests covering those boundaries. The updated exact-host guard and 48 focused URL/tab tests pass, but they do not establish safe isolation for these remaining cases.

ZacharyZcR:
Maintainer repair integrates current dev and adds an isolation boundary for embedded endpoints:

- Set `credentialless` before assigning the iframe URL, with an opaque sandbox origin and no popup/top-navigation permissions. This covers redirects as well as the initial hostname.
- Refuse unsupported browsers before creating a tunnel or frame.
- Disable external opening, which cannot preserve isolation in the user's shared browser cookie jar. New endpoints default to embedded mode; the UI explains the limitation.

Validation: 165 endpoint tests passed, followed by 38 UI/API/editor tests after the final UI changes. Type-check, lint (0 errors; existing warnings), and production build passed. A bounded real Chromium check followed a localhost redirect back to the parent hostname: neither endpoint request carried the parent cookie, the target could not access the parent DOM, and its Set-Cookie did not change the parent's session.

Compatibility limitation: pages requiring origin storage or popup-based sign-in may not function in this isolated mode. This is a security-first repair, not approval to merge with full feature compatibility claimed. Final review and CI are still required.

ZacharyZcR:
Maintainer optimization completed at `51c665a382c0d9c11a5ecf89b400c135409a5705`. Both PR Check jobs passed on this exact head; the PR is conflict-free against `dev-2.8.0`. The updated description records the repairs, verification and remaining runtime/compatibility limits. This PR has not been merged.

ZacharyZcR (approved):
Re-reviewed at 51c665a382c0d9c11a5ecf89b400c135409a5705. The previously reported cookie isolation blockers are resolved through credentialless sandboxed embedding and isolated, nonpersistent Electron sessions with no preload or Node access. Focused regression coverage (182 tests), the bounded native Electron isolation smoke check, and both PR CI jobs passed. The current head includes dev-2.8.0 and merges without conflicts. Approving this reviewed head.

PR #1418, opened 2026-09-09
Fix nested host references in remote sync

ZacharyZcR:
## Summary
Nested hosts must retain their parent relationship when desktop and server database IDs differ, including when an existing child is moved beneath a newly created parent on either side.

- Serialize parent references as stable sync IDs, resolve them within the authenticated user boundary, and validate parent relationships before writing.
- Order the actual cross-side writes by parent dependency for each destination, in addition to ordering individual responses. Reject cyclic planned dependencies before sending writes.
- Preserve timestamp conflict resolution and tombstone handling.

## Validation
- 9 focused reference/sync tests passed, including bidirectional reparenting, repeat synchronization, cyclic dependencies, and destination isolation.
- `npm run type-check`, `npm run lint`, `npm run build` passed.
- Integrated current `dev-2.8.0` without rewriting branch history.

Fixes Termix-SSH/Support#1275

### DIFF
--- a/src/backend/database/routes/sync-references.ts
+++ b/src/backend/database/routes/sync-references.ts
@@ -1,6 +1,6 @@
 import type { SyncEntityType } from "../repositories/sync-tombstone-repository.js";
 
-export type SyncReferenceEntity = "sshCredentials" | "vaultProfiles";
+export type SyncReferenceEntity = "hosts" | "sshCredentials" | "vaultProfiles";
 
 interface SyncReference {
   field: string;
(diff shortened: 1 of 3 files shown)


ZacharyZcR:
Maintainer audit: changes required before merge (author-account limitation prevents a formal Request Changes review).

Confirmed dependency-order bug in nested-host sync. `orderSyncRows` sorts each endpoint response, but `electron/remote-sync.cjs` constructs `allSyncIds` from local rows before remote rows. If local already has a child and remote creates a new parent and reparents that child, the child is processed before the new remote parent.

Reproduced using the actual sync engine method and the new reference deserializer: local=[old child], remote=[new parent, newer child referencing parent]. Sync throws `Missing hosts dependency parentHostSyncId=parent` before copying the parent; retrying fails at the same point.

Please order the actual planned cross-side writes by host dependencies, not only each response, and add integration tests for reparenting existing children in both directions. Keep this PR unmerged until this scenario passes.

PR #1420, opened 2026-09-10
fix(auth): skip TOTP only after verified WebAuthn user verification

neooriginal:
## Summary
- Treat a verified WebAuthn/security-key assertion as sufficient 2FA.
- Issue a full session on passkey login instead of prompting for TOTP when it is also enabled.
- Password login is unchanged: TOTP is still required when it is enabled.

## Test plan
- [x] Log in with a passkey on an account that also has TOTP enabled and confirm no TOTP prompt appears.
- [x] Log in with username/password on the same account and confirm TOTP is still required.
- [x] Confirm passkey login still works on an account without TOTP.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/backend/database/routes/user-webauthn-routes.ts
+++ b/src/backend/database/routes/user-webauthn-routes.ts
@@ -17,11 +17,7 @@ import { nanoid } from "nanoid";
 import type { AuthenticatedRequest } from "../../../types/index.js";
 import { AuthManager } from "../../utils/auth-manager.js";
 import { authLogger } from "../../utils/logger.js";
-import {
-  generateDeviceFingerprint,
-  getDeviceId,
-  parseUserAgent,
-} from "../../utils/user-agent-parser.js";
+import { parseUserAgent } from "../../utils/user-agent-parser.js";
 import {
   createCurrentUserRepository,
   createCurrentWebauthnCredentialRepository,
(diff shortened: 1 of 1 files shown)


ZacharyZcR (changes requested):
Request changes: TOTP is bypassed without requiring verified WebAuthn user verification.

The options endpoint accepts client-provided `userVerification` (`discouraged`, `preferred`, or `required`). Assertion verification sets `requireUserVerification` only when that requested value is `required`. This PR then unconditionally removes the `totpEnabled`/pending-TOTP branch and issues a full session after any verified assertion.

A valid possession-only security-key assertion with no PIN/biometric user verification can therefore bypass TOTP configured on the account. A verified signature/user-presence touch is not evidence that a second factor was verified.

Please retain the TOTP challenge unless server policy and the verified assertion establish the required user verification. Add route regression tests for a TOTP-enabled, untrusted device with UV absent/false, verified UV, and a client request using `userVerification: discouraged`. CI currently does not cover this authorization-policy change.

ZacharyZcR:
Maintainer repair pushed at `94e640ef9a92c247fe25ad6d7e1c2f6c82b6d8c9`. Original contributor history is preserved; current `dev-2.8.0` is integrated.

Only a verified WebAuthn assertion with `authenticationInfo.userVerified === true` can satisfy the configured TOTP requirement. Possession-only assertions retain TOTP and trusted-device handling. Added route regressions for all user-verification preferences, verified UV, trusted devices, invalid assertions and accounts without TOTP.

Validation: 7 focused tests passed. `npm run type-check`, `npm run lint` (0 errors; existing warnings), and `npm run build` passed locally. CI is being approved for the new head. This update does not mark the PR approved or merged.

ZacharyZcR:
Follow-up: CI found a Prettier normalization issue in the new regression test. Fixed and pushed at `8653cdfe`; the focused format check now passes. The previous head's Postgres/MySQL job passed, and CI has been approved again for this formatting-only follow-up.

ZacharyZcR (approved):
Re-reviewed repaired head 8653cdfe71daf7c248d0881bd5c9c76664b73006. The previously reported blockers are resolved. Both PR Check jobs pass on this exact head. The combined ten-PR integration tree passed 215 focused tests, type-check, lint, Prettier and production build. Approving the scoped change after checking its dependency delta and current dev merge.

PR #1430, opened 2026-09-15
fix(metrics): support macOS/WIN hosts in stats collection

neooriginal:
Metrics collection only ever ran Linux commands (/proc/stat, /proc/meminfo, /proc/uptime, /proc/net/dev, ip, GNU-only df flags). Any macOS host always came back with null/0% for cpu, memory, uptime, network and disk regardless of the enabled toggle in the webui.

Now detects the host platform once per poll and uses vm_stat, sysctl, top, ifconfig and netstat on Darwin hosts. Disk collection was also switched to a df flag combo that works on both GNU and BSD df.

Added tests for the new macOS parsers, existing ones untouched and passing.


neooriginal:
ready to be merged

ZacharyZcR (changes requested):
Request changes: the advertised macOS disk collection still executes GNU df options and returns no disk data.

`collectDiskMetrics(..., "darwin")` has no Darwin branch; only Windows is dispatched separately. It executes `df -hT -P` and `df -TkP`, then parses the GNU filesystem-type column layout.

Reproduced by running the actual collector with its command executor backed by the local macOS df: ordinary `df -k /` returns the root filesystem, while this collector returns an empty `filesystems` array and null disk metrics. macOS df does not use the GNU `-T` output format assumed here.

Please add a Darwin-compatible command/parser path and fixtures from actual macOS df output, covering root disk metrics and monitored/excluded mounts. The CPU/network additions and passing Linux CI do not validate this path.

neooriginal:
Good catch, fixed. BSD df's -T takes a filesystem type to filter by, not print - completely different from GNU df. macOS now goes through df -Pk + mount to build the same filesystem/type/percent list, skipping devfs/autofs and the synthetic APFS system volumes (Preboot, VM, Update, etc). Added parser tests using real df -Pk/mount output shapes.

Tested this on my own machine (M4, latest macOS) and disk usage reports correctly now, along with cpu/memory/network/uptime from the earlier commits.

neooriginal:
Bildschirmfoto 2026-09-15 um 23 25 03
proof

PR #1446, opened 2026-09-17
feat(file-manager): parallel local transfers

maxser0v:
## What

Uploads from the Local pane and downloads into it ran strictly one file after another, so batches of many small files were bound by per-file latency rather than bandwidth. This makes both directions run through a bounded worker pool.

- `runWithConcurrency(items, limit, worker, shouldStop)` in `local-transfer-utils.ts`: dispatches in order, keeps at most `limit` transfers in flight, stops dispatching once cancelled while letting in-flight items finish; a failing item never aborts the batch.
- `useLocalTransfers`: both directions use the pool. Directory skeletons are still created up front (so file order does not matter). The batch runner tracks the set of in-flight transfer ids — **Cancel** cancels all of them — sums bytes across in-flight + settled transfers for the progress bar / speed, and the toast shows how many files are moving at once ("4 at once").
- New setting **Simultaneous File Transfers** (1–8, default 4) in the profile panel → Interface, stored in `localStorage` (`termix:file-manager:transfer-concurrency`) and included in the storage snapshot. `1` restores the previous sequential behaviour; the hint suggests lowering it for servers that limit SFTP channels.
- Remote directories are created with the idempotent mkdir from #1442 without swallowing its errors, so a real failure stops the batch up front instead of surfacing as N per-file failures. When files do fail, the summary toast now carries the first underlying reason instead of a bare "Upload failed".

## Why it is safe on the backend

Each SSH session already shares one SFTP channel (`getSessionSftp` / `ChannelOpenSerializer`), so concurrent transfers reuse it and do not open extra SSH connections. Progress reporting stays per-transfer (150 ms throttle), so 8 concurrent files do not flood IPC.

## Tests

- Pool: never exceeds the limit and drains everything; stops on cancel; tolerates a failing item; handles empty input. Preference clamping and persistence.
- Bridge: 5 concurrent uploads keep every body intact and cancelling one leaves the others alone; 5 concurrent downloads with one cancelled mid-flight leave no `.termix-part` behind and every other file byte-identical.
- Manually: 6 concurrent uploads and 4 concurrent downloads through the real bridge against a real backend + OpenSSH server, all byte-identical.

Independent of the other fix PRs I'm opening today; it applies cleanly on `dev-2.8.0` on its own.


LukeGus:
Just finished reviewing all 6 of the PR's you opened. LGTM, thanks. I'll merge them all for v2.8.0.

PR #1452, opened 2026-09-18
feat(host-metrics): add NVIDIA GPU metrics card

Adelzu:
# Overview

Adds an opt-in NVIDIA GPU card to Host Metrics, collected over SSH with `nvidia-smi`.

- [x] Added: GPU collector and a GPU card with per-GPU utilization, VRAM, temperature, power, fan speed, live sparklines and the processes using each GPU
- [x] Updated: adaptive polling now also watches GPU readings, so it no longer backs off on hosts where only the GPU is busy
- [x] Fixed: every Host Metrics sparkline (CPU, memory, disk, GPU) stayed empty until the second poll, about 35 s after opening the tab. They now draw from the first sample.

# Changes Made

- **Collector** (`widgets/gpu-collector.ts`): one SSH exec per poll. It exits immediately when `nvidia-smi` is missing, then runs `--query-gpu` and `--query-compute-apps` in CSV form. Free-text names are queried last so commas in them can't shift the numeric columns, `[N/A]` / `[Not Supported]` become `null`, and processes are matched to GPUs by UUID. It is called from `collectMetrics()` next to the temperature collector, with the same try/catch fallback.
- **Data shape**: `{ source: "nvidia-smi" | "none", gpus, processes }`. Nothing in it is NVIDIA-specific, so AMD or Intel sources can be added later.
- **Types**: `"gpu"` added to `WidgetType` and `METRIC_CARD_IDS` (default width 2 columns), plus `ServerMetrics.gpu`.
- **UI**: `GpuCard.tsx`. Per-GPU sparkline history lives in `HostMetricsTab` through `appendGpuHistories` (`gpu-history.ts`).
- **Opt-in**: the card is in the Add card tray but not in new-host defaults, `DEFAULT_STATS_CONFIG` or any UI preset, since most servers have no GPU.
- **Polling signature**: `metricsChangeKey` moved unchanged from `HostMetricsTab.tsx` into `metrics-change-key.ts` so it can be tested, then extended with GPU utilization/VRAM (5% buckets, like the others) and the GPU process count.
- **Sparkline first sample** (separate `fix` commit): the tab's first fetch set the metrics but never added the sample to the sparkline history, so a sparkline (it needs 2 points) stayed empty until the next poll, one full interval later. The first sample is now recorded like every later one. Measured in the browser, the lines appeared at about 37 s before this change and about 3 s after it.
- **Strings**: `en.json` only (other languages come through Crowdin). The README feature list now mentions NVIDIA GPU.

**Limitations**

- Linux hosts only for now. On macOS and Windows hosts the collector returns `source: "none"`, and the command fails fast there. Windows NVIDIA hosts could be supported later with a PowerShell variant.
- `--query-compute-apps` lists compute (CUDA) processes, not graphics processes.
- Owners who add the card get `"gpu"` in `statsConfig.enabledWidgets`. The mobile app should ignore widget ids it doesn't know; worth a quick check on that side.

**Testing**

- New tests: `gpu-collector.test.ts` (N/A fields, commas in names, driver error text on stdout, header rows, empty output), `gpu-history.test.ts`, `GpuCard.test.tsx`, `metrics-change-key.test.ts`, `HostMetricsTab.test.tsx` (the first test for this tab: the trend lines draw from the first sample), plus a `host-metrics-preferences.test.ts` case for the new card id.
- `npm run lint`: 0 errors.
- `npm run type-check`: the only errors are the 6 in `file-manager/hooks/useLocalTransfers.ts`, which also fail on `dev-2.8.0` without this change.
- `vitest`: everything passes except `tunnel-action-authorization` and `web-endpoint-tunnel-manager`. Each hits the 5 s timeout under full-suite load on my machine and passes when run alone. Neither file is touched here.
- Tried in the running app against a remote Linux host with an NVIDIA GPU (RTX 2070 SUPER, driver 580), on desktop and at phone width.

# Related Issues

- None found. There is no existing issue for GPU metrics in Termix or Support.

# Screenshots / Demos

termix-gpu

# Checklist


LukeGus:
@copilot Fix the code for all comments in [this review thread](https://github.com/Termix-SSH/Termix/pull/1452#pullrequestreview-5247571004).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

LukeGus:
LGTM, thanks.

PR #1454, opened 2026-09-19
fix(macos): skip single-instance lock in Mac App Store builds

datlt4:
Under App Sandbox, Chromium's process singleton binds its Unix socket at /var/folders/<hash>/T/<bundle-id>/S/SingletonSocket, outside anything the sandbox grants the app write access to. The bind fails with EPERM:

  ERROR:chrome/browser/process_singleton_posix.cc:1179] Failed to bind()
  /var/folders/.../T/com.karmaa.termix/S/SingletonSocket:
  Operation not permitted (1)
  Another instance is already running, quitting...

requestSingleInstanceLock() treats that failure as a held lock, so the MAS build quits immediately on every launch and the app can never be opened.

macOS already reactivates the running instance when a bundled app is launched again, so gate the lock on process.mas and let the OS handle it there. Non-MAS builds are unchanged.

# Overview

- [ ] Added: ...
- [ ] Updated: ...
- [ ] Removed: ...
- [ ] Fixed: ...

# Changes Made

_Detailed explanation of changes (if needed)_

- ...

# Related Issues

_Link any issues this PR addresses_

- Closes #ISSUE_NUMBER
- Related to #ISSUE_NUMBER

# Screenshots / Demos

_(Optional: add before/after screenshots, GIFs, or console output)_

# Checklist

- [ ] Code follows project style guidelines
- [ ] I have read [Contributing.md](https://github.com/Termix-SSH/Termix/blob/main/CONTRIBUTING.md)
- [ ] This is not a translation request. See [docs](https://docs.termix.site/translations)


LukeGus:
LGTM, thanks

